PR #23, opened 2020-11-11
Feature/Watchers

georgiyazovaliiski:


app/process/model/essential/essential.go
```diff
@@ -0,0 +1,9 @@
+package essential
+
+type (
+	Essential struct {
```
failfmi, line 4: CryptoTransferMessage?

app/process/watcher/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,53 @@
...
+import (
+	"encoding/json"
+	hederasdk "github.com/hashgraph/hedera-sdk-go"
+	"github.com/limechain/hedera-eth-bridge-validator/config"
+	"github.com/limechain/hedera-watcher-sdk/queue"
+	"github.com/limechain/hedera-watcher-sdk/types"
+	"log"
+)
+
+type ConsensusTopicWatcher struct {
```
failfmi, line 12: Add NewConsensusTopicWatcher, which takes a HederaMirrorClient, topicId and typeMessage (to which the watcher will push messages) as parameters and returns *ConsensusTopicWater

app/process/watcher/crypto-transfer/crypto-transfer.go
```diff
@@ -1,3 +1,90 @@
...
+	"github.com/limechain/hedera-watcher-sdk/queue"
+	"github.com/limechain/hedera-watcher-sdk/types"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"strconv"
+	"time"
+)
+
+type CryptoTransferWatcher struct {
```
failfmi, line 19: Add NewCryptoTransferWatcher, which takes httpClient, accountId and  typeMessage(to which the watcher will push messages ) as parameters and returns a *CryptoTransferWatcher

app/process/watcher/crypto-transfer/crypto-transfer.go
```diff
@@ -1,3 +1,90 @@
...
+
+func getTransactionsFor(account hederasdk.AccountID, lastProcessedTimestamp string) (*transaction.Transactions, error) {
+	// TODO: Get last processed Tx timestamp
+	address := fmt.Sprintf("%s%s", config.LoadConfig().Hedera.MirrorNode.ApiAddress, "transactions")
+	accountLink := fmt.Sprintf("%s?account.id=%s&type=credit&result=success&timestamp=gt:%s&order=asc",
+		address,
+		account.String(),
+		lastProcessedTimestamp)
+
+	response, e := http.Get(accountLink)
```
failfmi, line 35: Let's extract the http calls to a separate client, e.g.
`httpClient.GetTransactionsByAccountIdAndTimestamp`

failfmi (changes requested):

cmd/main.go
```diff
@@ -2,13 +2,33 @@ package main
...
+
+		server.AddWatcher(crypto_transfer.CryptoTransferWatcher{Account: id})
+	}
+	for _, topic := range configuration.Hedera.Watcher.ConsensusMessage.Topics {
+		id, e := hedera.TopicIDFromString(topic.Id)
+		if e != nil {
+			panic(e)
+		}
+
+		server.AddWatcher(consensus_message.ConsensusTopicWatcher{TopicID: id})
```
failfmi, line 31: NewConsensusTopicWatcher(mirrorClient, id, messageType)

cmd/main.go
```diff
@@ -2,13 +2,33 @@ package main
...
 	persistence.RunDb(configuration.Hedera.Validator.Db)
-	server.NewServer().Run(fmt.Sprintf(":%s", configuration.Hedera.Validator.Port))
+	server := server.NewServer()
+	for _, account := range configuration.Hedera.Watcher.CryptoTransfer.Accounts {
+		id, e := hedera.AccountIDFromString(account.Id)
+		if e != nil {
+			panic(e)
+		}
+
+		server.AddWatcher(crypto_transfer.CryptoTransferWatcher{Account: id})
```
failfmi, line 23: NewCryptoTransferWatcher(httpClient, id, messageType)

cmd/main.go
```diff
@@ -2,13 +2,39 @@ package main
...
 	"github.com/limechain/hedera-watcher-sdk/server"
 )
 
 func main() {
 	configuration := config.LoadConfig()
 	persistence.RunDb(configuration.Hedera.Validator.Db)
-	server.NewServer().Run(fmt.Sprintf(":%s", configuration.Hedera.Validator.Port))
+	server := server.NewServer()
+
+	mirrorNodeClient, _ := hederasdk.NewMirrorClient(configuration.Hedera.MirrorNode)
```
failfmi, line 20: I think that error should be checked

app/process/watcher/crypto-transfer/crypto-transfer.go
```diff
@@ -1,3 +1,61 @@
...
+				information := crypto_transfer_message.CryptoTransferMessage{
+					TxMemo: tx.MemoBase64,
+					Sender: tx.Transfers[len(tx.Transfers)-2].Account,
+					Amount: tx.Transfers[len(tx.Transfers)-1].Amount,
+				}
+				proceed.Proceed(information, typeMessage, account, q)
+			}
+			lastObservedTimestamp = transactions.Transactions[len(transactions.Transactions)-1].ConsensusTimestamp
+		}
+		time.Sleep(5 * time.Second)
```
failfmi, line 51: I think that the frequency of this call should be extracted into a configuration parameter. We can discuss it.

cmd/main.go
```diff
@@ -2,13 +2,45 @@ package main
...
+	"log"
 )
 
 func main() {
 	configuration := config.LoadConfig()
 	persistence.RunDb(configuration.Hedera.Validator.Db)
-	server.NewServer().Run(fmt.Sprintf(":%s", configuration.Hedera.Validator.Port))
+	server := server.NewServer()
+
+	mirrorNodeClient, e := hederasdk.NewMirrorClient(configuration.Hedera.MirrorNode)
```
Daniel-K-Ivanov, line 21: We should be creating a new `hedera` client and not use the `hedera.NewMirrorClient`. I would actually think about creating the instance of the client inside the watchers and not globally here. At some point the client might have a PK to sign transactions. In that case, we would like to create an instance of the hedera client with a given PK and that client would be used only by the cryptotransfer handler that should post consensus messages

cmd/main.go
```diff
@@ -2,13 +2,45 @@ package main
...
+
+	httpClient := http.NewClient(configuration.Hedera.MirrorNode.ApiAddress)
+
+	for _, account := range configuration.Hedera.Watcher.CryptoTransfer.Accounts {
+		id, e := hedera.AccountIDFromString(account.Id)
+		if e != nil {
+			panic(e)
+		}
+
+		server.AddWatcher(crypto_transfer.NewCryptoTransferWatcher(httpClient, id, configuration.Hedera.MirrorNode.PollingInterval))
```
Daniel-K-Ivanov, line 35: I think that it would be good to have loggers here stating that there were N crypto watchers added and if no watchers were added we should log that as-well

cmd/main.go
```diff
@@ -2,13 +2,45 @@ package main
...
+	}
+	for _, topic := range configuration.Hedera.Watcher.ConsensusMessage.Topics {
+		id, e := hedera.TopicIDFromString(topic.Id)
+		if e != nil {
+			panic(e)
+		}
+
+		server.AddWatcher(consensus_message.NewConsensusTopicWatcher(mirrorNodeClient, id))
```
Daniel-K-Ivanov, line 43: I think that it would be good to have loggers here stating that there were N topic watchers added and if no watchers were added we should log that as-well

config/application.yml
```diff
@@ -7,4 +7,24 @@ hedera:
...
+    polling_interval: 5000000000
+  client:
+    network_type: testnet
+    operator:
+      account_id:
+      private_key:
+  watcher:
+    crypto-transfer:
+      accounts:
+#        - id: shard.realm.account
```
Daniel-K-Ivanov, line 23: I think that we should remove the commented lines here

Daniel-K-Ivanov (changes requested):

app/clients/hedera/hedera.go
```diff
@@ -1,3 +1,58 @@
...
+	"encoding/json"
+	"fmt"
+	hederasdk "github.com/hashgraph/hedera-sdk-go"
+	"github.com/limechain/hedera-eth-bridge-validator/app/process/model/transaction"
+	"io/ioutil"
+	"log"
+	"net/http"
+)
+
+type Client struct {
```
Perseverance, line 13: `Client` is a bit too generic for me. I'd name this `HederaClient`

app/clients/hedera/hedera.go
```diff
@@ -1,3 +1,58 @@
...
+	mirrorAPIAddress string
+	mirrorClient     *hederasdk.MirrorClient
+	httpClient       *http.Client
+}
+
+func (hedera Client) GetMirror() *hederasdk.MirrorClient {
```
Perseverance, line 19: I'd not name the variable `hedera` a bit ambiguous - either go with the `client` or even short `c`

Perseverance, line 19: GetMirror is a bit like a fairytale to me :D Lets add the word client at the end - `GetMirrorClient`

app/clients/hedera/hedera.go
```diff
@@ -1,3 +1,58 @@
...
+	return hedera.mirrorClient
+}
+
+func NewClient(mirrorAPIAddress, mirrorNodeClientAddress string) *Client {
```
Perseverance, line 23: When you rename the name of the Struct please change the name here too

Perseverance, line 23: In the two parameters, you use once just `mirror` and once `mirrorNode`. I'd preffert to stick to `mirrorNode` in both occasions.

app/clients/hedera/hedera.go
```diff
@@ -1,3 +1,58 @@
...
+	mirrorClient, e := hederasdk.NewMirrorClient(mirrorNodeClientAddress)
+	if e != nil {
+		log.Printf("Error: Could not instantiate mirror client - [%s]", e)
```
Perseverance, line 26: I would `panic` here. If there is no client, no need for me to continue.

Perseverance: Alternatively, will return an error and let the creator deal with it

app/clients/hedera/hedera.go
```diff
@@ -1,3 +1,58 @@
...
+	}
+
+	return &Client{
+		mirrorAPIAddress: mirrorAPIAddress,
+		mirrorClient:     &mirrorClient,
+		httpClient:       &http.Client{},
+	}
+}
+
+func (hedera Client) GetTransactionsByAccountIdAndTimestamp(account hederasdk.AccountID, lastProcessedTimestamp string) (*transaction.Transactions, error) {
```
Perseverance, line 36: Just the name of the transaction does not imply that you are going to get the transactions after a certain timestamp. I would name this `GetAccountTransactionsAfterDate` or something similar

Perseverance, line 36: I think the `account` parameter should be named `accountId`.
I think the `lastProcessedTimestamp` parameter is very context-specific name. I would name it `milestoneTimestamp` or something indicating a specific point in time.

Perseverance: As timestamps are numbers, I would expect the timestamp parameter to be some kind of `unit`, not `string`. This way we are forcing conversion for the person using the function

app/clients/hedera/hedera.go
```diff
@@ -1,3 +1,58 @@
...
+	address := fmt.Sprintf("%s%s", hedera.mirrorAPIAddress, "transactions")
+	accountLink := fmt.Sprintf("%s?account.id=%s&type=credit&result=success&timestamp=gt:%s&order=asc",
```
Perseverance, line 38: A bit too hardcoded for me, but cant think of a way to make it better - @Daniel-K-Ivanov @failfmi any thoughts?

app/clients/hedera/hedera.go
Perseverance, line 36: I think the idiomatic go places the error as the first return parameter, and the result as the second

app/clients/hedera/hedera.go
Perseverance, line 38: Is this really an `accountLink` or some kind of transactions download url.

app/process/model/transaction/transaction.go
```diff
@@ -0,0 +1,25 @@
+package transaction
+
+type (
+	Transaction struct {
```
Perseverance, line 4: I wonder if placing `Hedera` in front of the models would be a good fit, enabling people reading realizing that this is Hedera specific. Ethereum transactions would have a much different format.

app/process/watcher/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,48 @@
-package consensus_message
+package consensusmessage
 
-// todo consensus message watcher
+import (
+	hederasdk "github.com/hashgraph/hedera-sdk-go"
```
Perseverance, line 4: I would leave the external hedera sdk to carry the name `hedera`, and give an alias to our `HederaClient` something like `hederaClient. The way that you've done it I keep getting confused if you are using something from the sdk or something we've written. cc @Daniel-K-Ivanov @failfmi

app/process/watcher/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,48 @@
...
+	"github.com/limechain/hedera-eth-bridge-validator/app/clients/hedera"
+	"github.com/limechain/hedera-eth-bridge-validator/app/process/watcher/publisher"
+	"github.com/limechain/hedera-watcher-sdk/queue"
+	"log"
```
Perseverance, line 8: Not for this PR but in general think about including a more powerful logging library that will allow you to do stuff like `log.Infof`, `log.Warnf` and indicate error levels. It will also allow you to disable certain level messages

app/process/watcher/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,48 @@
...
+	_, e := hederasdk.NewMirrorConsensusTopicQuery().
+		SetTopicID(topicId).
+		Subscribe(
+			*ctw.client.GetMirror(),
+			func(response hederasdk.MirrorConsensusTopicResponse) {
+				log.Printf("[%s] - Topic [%s] - Response incoming: [%s]", response.ConsensusTimestamp, topicId, response.Message)
+				publisher.Publish(response, typeMessage, topicId, q)
+			},
+			func(err error) {
+				log.Printf("Error incoming: [%s]", err)
```
Perseverance, line 31: We need to figure out our recovery policy. Mirror nodes tend to disconnect themselves and we should be able to reconnect. cc @Daniel-K-Ivanov @failfmi

app/process/watcher/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,48 @@
...
+			},
+		)
+
+	if e != nil {
+		log.Printf("Did not subscribe to [%s].", topicId)
```
Perseverance, line 36: I feel if there is an error, it should be told to the subscriber.

app/process/watcher/crypto-transfer/crypto-transfer.go
```diff
@@ -1,3 +1,63 @@
...
+	typeMessage     string
+	pollingInterval time.Duration
+}
+
+func (ctw CryptoTransferWatcher) Watch(queue *queue.Queue) {
+	go ctw.beginWatching(ctw.accountID, ctw.typeMessage, queue)
+}
+
+func (ctw CryptoTransferWatcher) beginWatching(account hederasdk.AccountID, typeMessage string, q *queue.Queue) {
+	lastObservedTimestamp := strconv.FormatInt(time.Now().Unix(), 10)
```
Perseverance, line 26: Is this something temporary, until we have Database. If so a comment would be good

app/process/watcher/crypto-transfer/crypto-transfer.go
```diff
@@ -1,3 +1,63 @@
...
+	for {
+		transactions, e := ctw.client.GetTransactionsByAccountIdAndTimestamp(account, lastObservedTimestamp)
+		if e != nil {
+			log.Printf("Suddenly stopped monitoring config [%s]\n", account.String())
```
Perseverance, line 30: I think the subscriber should do with the errors

Perseverance (changes requested):

app/process/watcher/crypto-transfer/crypto-transfer.go
georgiyazovaliiski: Yes. It is temporary

app/clients/hedera/hedera.go
Perseverance: Ignore this, I was drunk!

Perseverance (changes requested):

Perseverance (approved):

app/persistence/status/status.go
```diff
@@ -1,5 +1,63 @@
...
+func (s StatusRepository) GetLastFetchedTimestamp(accountID hedera.AccountID) string {
+	lastFetchedStatus := &Status{}
+	failure := s.dbClient.
+		Table("statuses").
+		Where("code = ? and account_id = ?", "LAST_FETCHED_TIMESTAMP", accountID.String()).
+		First(&lastFetchedStatus).Error
+
+	if failure != nil && errors.Is(failure, gorm.ErrRecordNotFound) {
+		log.Errorf("Could not get last fetched timestamp: [%s]\n", failure)
+		now := time.Now()
```
Daniel-K-Ivanov, line 39: We shouldn't be setting "last fetched timestamp" before we actually fetched it. You should return `nil` or error and handle that in the watcher.

app/persistence/status/status.go
```diff
@@ -1,5 +1,63 @@
...
+		dbClient: dbClient,
+	}
+}
+
+func (s StatusRepository) GetLastFetchedTimestamp(accountID hedera.AccountID) string {
+	lastFetchedStatus := &Status{}
+	failure := s.dbClient.
+		Table("statuses").
+		Where("code = ? and account_id = ?", "LAST_FETCHED_TIMESTAMP", accountID.String()).
+		First(&lastFetchedStatus).Error
```
Daniel-K-Ivanov, line 35: You can do directly `.RecordNotFound()` in an if statement

app/persistence/status/status.go
```diff
@@ -1,5 +1,63 @@
...
+func NewStatusRepository(dbClient *gorm.DB) *StatusRepository {
+	return &StatusRepository{
+		dbClient: dbClient,
+	}
+}
+
+func (s StatusRepository) GetLastFetchedTimestamp(accountID hedera.AccountID) string {
+	lastFetchedStatus := &Status{}
+	failure := s.dbClient.
+		Table("statuses").
```
Daniel-K-Ivanov, line 33: I don't think that you need to define the `table` at all. You can use the `Status`

app/persistence/status/status.go
```diff
@@ -1,5 +1,63 @@
...
+	"strconv"
+	"time"
+)
+
 // This table will contain information for latest status of the application
 type Status struct {
+	Name      string
+	AccountID string
+	Code      string
+	Value     string
```
Daniel-K-Ivanov, line 17: `Value` is not that self-explanatory. 
Personally, I think that the DB should have
two columns - `account_id` and `last_fetched_timestamp`
which are `string` and `int64`

Are we going to store something else in there other than the last fetched timestamp for an address?

app/persistence/status/status.go
```diff
@@ -1,5 +1,63 @@
...
+	return lastFetchedStatus.Value
+}
+
+func (s StatusRepository) UpdateLastFetchedTimestamp(accountID hedera.AccountID, timestamp string) error {
+	return s.dbClient.
+		Where("code = ? and account_id = ?", "LAST_FETCHED_TIMESTAMP", accountID.String()).
+		Save(Status{
+			Name:      "Last fetched timestamp",
+			AccountID: accountID.String(),
+			Code:      "LAST_FETCHED_TIMESTAMP",
```
Daniel-K-Ivanov, line 59: You are using `LAST_FETCHED_TIMESTAMP` a lot. Think about introducing it as const variable at the top.

app/process/watcher/crypto-transfer/crypto-transfer.go
```diff
@@ -1,3 +1,88 @@
...
+			for _, tx := range transactions.Transactions {
+				log.Infof("[%s] - New transaction on account [%s] - Tx Hash: [%s]\n",
+					tx.ConsensusTimestamp,
+					ctw.accountID.String(),
+					tx.TransactionHash)
+
+				information := cryptotransfermessage.CryptoTransferMessage{
+					TxMemo: tx.MemoBase64,
+					Sender: tx.Transfers[len(tx.Transfers)-2].Account,
+					Amount: tx.Transfers[len(tx.Transfers)-1].Amount,
```
Daniel-K-Ivanov, line 63: I don't think that the indices -1 and -2 should be hardcoded here. What if there are 10 transfers in the transaction? will this hold?

app/process/watcher/crypto-transfer/crypto-transfer.go
```diff
@@ -1,3 +1,88 @@
...
+				}
+				publisher.Publish(information, ctw.typeMessage, ctw.accountID, q)
+			}
+			milestoneTimestamp = transactions.Transactions[len(transactions.Transactions)-1].ConsensusTimestamp
+		}
+
+		failure := ctw.statusRepository.UpdateLastFetchedTimestamp(ctw.accountID, milestoneTimestamp)
```
Daniel-K-Ivanov, line 70: normally in `go` you have `err` not failure

app/process/watcher/crypto-transfer/crypto-transfer.go
```diff
@@ -1,3 +1,88 @@
...
+		if failure != nil {
+			log.Errorf("Error incoming: Suddenly stopped monitoring account [%s]\n", ctw.accountID.String())
```
Daniel-K-Ivanov, line 72: If we were not able to `updateLastFetchedTimestamp` we should not log only that we are `suddenly stopped monitoring account`, but we need to state that db update failed

app/process/watcher/publisher/publish.go
```diff
@@ -0,0 +1,20 @@
...
+	"encoding/json"
+	"github.com/limechain/hedera-watcher-sdk/queue"
+	"github.com/limechain/hedera-watcher-sdk/types"
+	log "github.com/sirupsen/logrus"
+)
+
+func Publish(m interface{}, typeMessage string, id interface{}, q *queue.Queue) {
+	message, e := json.Marshal(m)
+	if e != nil {
+		log.Errorf("[%s] - Failed marshalling response - ID: [%s]\n", typeMessage, id)
```
Daniel-K-Ivanov, line 13: if marshalling fails, we should not try to do `q.Push`

cmd/main.go
```diff
@@ -1,14 +1,75 @@
...
 
 func main() {
+	initLogger()
 	configuration := config.LoadConfig()
-	persistence.RunDb(configuration.Hedera.Validator.Db)
-	server.NewServer().Run(fmt.Sprintf(":%s", configuration.Hedera.Validator.Port))
+	hederaClient := hederaClient.NewHederaClient(configuration.Hedera.MirrorNode.ApiAddress, configuration.Hedera.MirrorNode.ClientAddress)
+	server := server.NewServer()
+
+	db := persistence.RunDb(configuration.Hedera.Validator.Db)
```
Daniel-K-Ivanov, line 24: When we run the db, we must `defer` the db connection

cmd/main.go
```diff
@@ -1,14 +1,75 @@
...
+	statusRepository := status.NewStatusRepository(db)
+
+	failure := addCryptoTransferWatchers(configuration, hederaClient, statusRepository, server)
```
Daniel-K-Ivanov, line 27: normally `failure` is not used, but `err`

cmd/main.go
```diff
@@ -1,14 +1,75 @@
...
+	if failure != nil {
+		log.Fatal(failure)
+	}
+
+	server.Run(fmt.Sprintf(":%s", configuration.Hedera.Validator.Port))
+}
+
+func addCryptoTransferWatchers(configuration *config.Config, hederaClient *hederaClient.HederaClient, repository *status.StatusRepository, server *server.HederaWatcherServer) error {
+	if len(configuration.Hedera.Watcher.CryptoTransfer.Accounts) == 0 {
+		log.Warningln("There are no Crypto Transfer Watchers.")
```
Daniel-K-Ivanov, line 42: `There are no Crypto Transfer Watchers.` -> `CryptoTransfer Accounts list is empty. No CryptoTransfer Watchers will be started`

cmd/main.go
```diff
@@ -1,14 +1,75 @@
...
+
+		server.AddWatcher(cryptotransfer.NewCryptoTransferWatcher(hederaClient, id, configuration.Hedera.MirrorNode.PollingInterval, repository, account.MaxRetries))
+		log.Infof("Added a Crypto Transfer Watcher for account [%s]\n", account.Id)
+	}
+	return nil
+}
+
+func addConsensusTopicWatchers(configuration *config.Config, hederaClient *hederaClient.HederaClient, server *server.HederaWatcherServer) error {
+	if len(configuration.Hedera.Watcher.ConsensusMessage.Topics) == 0 {
+		log.Warningln("There are no Consensus Topic Watchers.")
```
Daniel-K-Ivanov, line 58: `There are no Consensus Topic Watchers.` -> `Consensus Message Topics list is empty. No Consensus Topic Watchers will be started`

Daniel-K-Ivanov (changes requested):

app/process/watcher/crypto-transfer/crypto-transfer.go
georgiyazovaliiski: This log is meant to be a description for the next line - log.Errorln(e).

app/process/watcher/crypto-transfer/crypto-transfer.go
georgiyazovaliiski: As far as I researched, when a crypto-transfer happens, the Receiver is on the second-to-last position and the Sender is on the last one, no matter how many transfers occur.

cmd/main.go
failfmi: No need to defer the db connection, the driver handles it itself.

app/process/watcher/crypto-transfer/crypto-transfer.go
failfmi: A new way to retrieve the data from the transfers will be implemented

app/process/model/transaction/transaction.go
```diff
@@ -0,0 +1,25 @@
...
+		Result               string `json:"result"`
+		Name                 string `json:"name"`
+		MaxFee               string `json:"max_fee"`
+		ValidDurationSeconds string `json:"valid_duration_seconds"`
+		Node                 string `json:"node"`
+		TransactionID        string `json:"transaction_id"`
+		Transfers            []Transfer
+	}
+	Transfer struct {
+		Account string `json:"config"`
```
failfmi, line 19: the json representation should be `account`

georgiyazovaliiski: The response from the Mirror API contains account id as a string.

app/process/watcher/consensus-message/consensus-message.go
```diff
@@ -1,42 +1,105 @@
...
+	}
+	return ctw.startTimestamp
+}
+
 func (ctw ConsensusTopicWatcher) subscribeToTopic(q *queue.Queue) {
-	_, e := hedera.NewMirrorConsensusTopicQuery().
+	var err error
+	milestoneTimestamp := ctw.startTimestamp
+
+	if !ctw.started {
```
Perseverance, line 56: In programing, there is the concept of "boobs". Boobs is when you have many nested ifs that can be untested by reversing the logic. Unlike real boobs, programming boobs are hard to read and look at and should be restricted and eliminated by reversing the logic.

An example of fixing this boobs issue is by making a method `initConsensusWatcher`.

Instead of nesting ifs, you can reverse the logic and return. Example:
```
func initConsensusWatcher(ctw ConsensusTopicWatcher) {
if milestoneTimestamp != "" {
return
}
log.Warnf("[%s] Starting Timestamp was empty, proceeding to get [timestamp] from database.", ctw.topicID.String())
milestoneTimestamp, err = ctw.statusRepository.GetLastFetchedTimestamp(ctw.topicID.String())
if err == nil || !errors.Is(err, gorm.ErrRecordNotFound) {
return
}
log.Warnf("[%s] Database Timestamp was empty, proceeding with [timestamp] from current moment.", ctw.topicID.String())
milestoneTimestamp = strconv.FormatInt(time.Now().Unix(), 10)
e := ctw.statusRepository.CreateTimestamp(ctw.topicID.String(), milestoneTimestamp)
if e != nil {
log.Fatal(e)
}
}
```

Sorry, a bit hard to write code in here

app/process/watcher/consensus-message/consensus-message.go
```diff
@@ -1,42 +1,105 @@
...
+				milestoneTimestamp = strconv.FormatInt(time.Now().Unix(), 10)
+				e := ctw.statusRepository.CreateTimestamp(ctw.topicID.String(), milestoneTimestamp)
+				if e != nil {
+					log.Fatal(e)
+				}
+			}
+		}
+	} else {
+		milestoneTimestamp, err = ctw.statusRepository.GetLastFetchedTimestamp(ctw.topicID.String())
+		if err != nil && errors.Is(err, gorm.ErrRecordNotFound) {
```
Perseverance, line 71: Same boobs comment here

app/process/watcher/crypto-transfer/crypto-transfer.go
```diff
@@ -38,9 +46,36 @@ func (ctw CryptoTransferWatcher) Watch(q *queue.Queue) {
...
 		log.Errorf("Error incoming: Could not start monitoring account [%s]\n", ctw.accountID.String())
+		return
 	}
+	log.Infof("Starting Crypto Transfer Watcher for account [%s]\n", ctw.accountID)
+
+	var err error
+	milestoneTimestamp := ctw.startTimestamp
+
+	if !ctw.started {
+		if milestoneTimestamp == "" {
```
Perseverance, line 57: Boobs ( . )( . ) (and also duplication? Oddly familiar code.)

Perseverance (changes requested):

failfmi (approved):
Let's just merge the PR, so we can unblock everyone needed, bugs will be fixed later.

Perseverance (approved):

Daniel-K-Ivanov (approved):

Daniel-K-Ivanov (approved):

PR #28, opened 2020-11-18
Feature: Crypto Transfer Event Handler

### LINKED ISSUE #27: CryptoTransfer Event Handler flow
Process for new TX:
0. Add new TX in the DB
3. Fee Calculation
4. Sign Authorisation Message
5. Submit to HCS
6. Update DB TX with the HCS TX Hash (containing the Signature)

failfmi:
**Detailed description**:
* Separate HederaMirrorNodeClient and HederaNodeClient
* Add EthSigner
* Add mocked FeeCalculator
* Add CryptoTransfer Event Handler flow
* Update protobuf marshal/unmarshal in watchers & handlers
* Switch fee to string due to possibility of overflowing uint64
* Fix improper parse of past topic messages.

Closes #27 
* Why tx completion is not done after submission? All nodes update state on mirror notification, so architecturally would be better to complete it later on, e.g. in the topic watcher.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


georgiyazovaliiski (approved):

app/process/handler/crypto-transfer/crypto-transfer.go
```diff
@@ -1,3 +1,113 @@
...
+	}
+
+	exists, err := cth.transactionRepo.Exists(ctm.TransactionId)
+	if err != nil {
+		log.Error(fmt.Sprintf("Error while trying to get transaction [%s] from database", ctm.TransactionId))
+		return err
+	}
+
+	if exists {
+		return nil
```
Daniel-K-Ivanov, line 40: it would be a good idea to log that we received new `Handle` for an already existing TX

app/process/handler/crypto-transfer/crypto-transfer.go
```diff
@@ -1,3 +1,113 @@
...
+	}
+
+	err = cth.transactionRepo.Create(&ctm)
+	if err != nil {
```
Daniel-K-Ivanov, line 44: I think that we should log in both cases:
- when we create the new transaction in the DB
- when we tried to create it but there was an error ( I know that there is a logger when we return this `err` but we won't have the context `when` the error occurred

app/process/handler/crypto-transfer/crypto-transfer.go
```diff
@@ -1,3 +1,113 @@
...
+	}
+
+	encodedSignature := hex.EncodeToString(signature)
+
+	topicMsgSubmissionTxId, err := cth.handleTopicSubmission(&ctm, encodedSignature)
+	if err != nil {
+		return err
+	}
+
+	return cth.transactionRepo.UpdateStatusSubmitted(ctm.TransactionId, topicMsgSubmissionTxId, encodedSignature)
```
Daniel-K-Ivanov, line 76: Even though we submitted the status and received the receipt (pre-check validations went through) we are not 100% sure that the transaction will succeed. I think that we should think about the use-case where we submit a tx, receive receipt, update the status to submitted and the tx fails.

app/process/handler/crypto-transfer/crypto-transfer.go
```diff
@@ -1,3 +1,113 @@
...
+	signature, err := cth.ethSigner.Sign(hash)
+	if err != nil {
+		return err
+	}
+
+	encodedSignature := hex.EncodeToString(signature)
+
+	topicMsgSubmissionTxId, err := cth.handleTopicSubmission(&ctm, encodedSignature)
+	if err != nil {
+		return err
```
Daniel-K-Ivanov, line 73: I think that we should log the context of the error (that topic submission failed for a given transaction). It is important to log the exact transaction ID

Daniel-K-Ivanov (changes requested):

app/process/handler/crypto-transfer/crypto-transfer.go
failfmi: This is the scenario I have explained as special notes in the PR above. To complete a tx, we should await the event from mirror node.

Daniel-K-Ivanov (approved):

PR #31, opened 2020-11-23
Feature / Consensus Message Handler

georgiyazovaliiski:


app/domain/repositories/message.go
```diff
@@ -0,0 +1,8 @@
+package repositories
+
+import "github.com/limechain/hedera-eth-bridge-validator/app/persistence/message"
+
+type MessageRepository interface {
+	Get(txId, signature string) ([]message.TransactionMessage, error)
```
failfmi, line 6: Good idea should be to explain what you get in the function name (GetByTxIdAndSignature).

app/persistence/message/message.go
```diff
@@ -0,0 +1,37 @@
+package message
+
+import (
+	"gorm.io/gorm"
+)
+
+type TransactionMessage struct {
```
failfmi, line 7: You can add the signer address / public key in case you need to check some stuff later.

app/persistence/message/message.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+func NewMessageRepository(dbClient *gorm.DB) *MessageRepository {
+	return &MessageRepository{
+		dbClient: dbClient,
+	}
+}
+
+func (m MessageRepository) Get(txId, signature string) ([]TransactionMessage, error) {
+	var signatures []TransactionMessage
```
failfmi, line 27: nit: rename to `messages`

app/persistence/message/message.go
```diff
@@ -0,0 +1,37 @@
...
+	err := m.dbClient.Where("transaction_id = ? and signature = ?", txId, signature).Find(&signatures).Error
+	if err != nil {
+		return nil, err
+	}
+	return signatures, nil
+}
+
+func (m MessageRepository) Add(message *TransactionMessage) error {
```
failfmi, line 35: It would be good if all repository creation/addition methods have the same name.

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,103 @@
...
+	decodedSig, err := hex.DecodeString(m.GetSignature())
+	if err != nil {
+		return errors.New(fmt.Sprintf("Failed to decode signature. - [%s]", err))
+	}
+
+	hash := crypto.Keccak256([]byte(fmt.Sprintf("%s-%s-%d-%s", ctm.TransactionId, ctm.EthAddress, ctm.Amount, ctm.Fee)))
+
+	log.Printf("[%s] hex hash", hex.EncodeToString(hash))
+
+	key, err := crypto.Ecrecover(hash, decodedSig)
```
failfmi, line 58: Missing error check.

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,103 @@
...
+	}
+
+	fmt.Println("Success.")
+	return nil
+}
+
+func isValidAddress(key string) bool {
+	keys := config.LoadConfig().Hedera.Handler.ConsensusMessage.Addresses
+	for _, k := range keys {
+		if k == key {
```
failfmi, line 98: You should check both strings `ToLower`

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,103 @@
...
+	if err != nil {
+		return errors.New(fmt.Sprintf("Could not add Transaction Message with Transaction Id and Signature - [%s]-[%s]", m.TransactionId, m.Signature))
+	}
+
+	fmt.Println("Success.")
+	return nil
+}
+
+func isValidAddress(key string) bool {
+	keys := config.LoadConfig().Hedera.Handler.ConsensusMessage.Addresses
```
failfmi, line 96: Save the addresses in Handler struct instead of loading the config every time.

app/process/watcher/consensus-message/consensus-message.go
```diff
@@ -88,15 +90,29 @@ func (ctw ConsensusTopicWatcher) subscribeToTopic(q *queue.Queue) {
...
+		if err != nil {
+			log.Errorf("Could not decode message - [%s]", u.Message)
+			log.Fatal(err)
+		}
+
+		msg := &validatorproto.TopicSignatureMessage{}
+		err = proto.Unmarshal(decodedMessage, msg)
+		if err != nil {
+			log.Errorf("Could not unmarshal message - [%s]", u.Message)
+			log.Fatal(err)
```
failfmi, line 109: It is not a good idea to fatal protobuf unmarshalling - if message of another type is submitted, it is still valid, just not for this scenario.

app/process/watcher/publisher/publish.go
```diff
@@ -12,7 +12,7 @@ func Publish(m proto.Message, typeMessage string, id interface{}, q *queue.Queue
 	if e != nil {
 		log.Fatalf("[%s] - Failed marshalling response - ID: [%s]\n", typeMessage, id)
 	}
-
+	log.Println(m)
```
failfmi, line 15: nit: unnecessary log

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,103 @@
...
+	"github.com/golang/protobuf/proto"
+	"github.com/limechain/hedera-eth-bridge-validator/app/domain/repositories"
+	"github.com/limechain/hedera-eth-bridge-validator/app/persistence/message"
+	"github.com/limechain/hedera-eth-bridge-validator/config"
+	validatorproto "github.com/limechain/hedera-eth-bridge-validator/proto"
+	log "github.com/sirupsen/logrus"
+)
+
+type ConsensusMessageHandler struct {
+	repository repositories.MessageRepository
```
failfmi, line 17: nit: `*repositories.MessageRepository`

failfmi (changes requested):

app/persistence/message/message.go
georgiyazovaliiski: Figured that out earlier today, when coming up with the election process. Thanks

cmd/main.go
```diff
@@ -34,6 +36,7 @@ func main() {
 
 	statusCryptoTransferRepository := status.NewStatusRepository(db, "CRYPTO_TRANSFER")
 	statusConsensusMessageRepository := status.NewStatusRepository(db, "HCS_TOPIC")
```
Daniel-K-Ivanov, line 38: We can move those 3 lines after the `transactionRepository` just so that we can have the initialisation of repositories together

cmd/main.go
```diff
@@ -45,6 +48,8 @@ func main() {
 		log.Fatal(err)
 	}
 
+	server.AddHandler("HCS_TOPIC_MSG", cmh.NewConsensusMessageHandler(*messageRepository))
```
Daniel-K-Ivanov, line 51: Let's keep the code consistent:
1. Create db, clients, repositories
2. Create server
3. Add CryptoTransfer handler + watcher
4. Add HCS handler + watcher

Daniel-K-Ivanov, line 51: One general feedback that I have is that the `messageTypes` of the handlers/watchers are hardcoded in different places. 
We ideally want to have them configured/hardcoded into one place only.
`HCS_TOPIC_MSG` is hardcoded here and in `consensus-message.go`.

One suggestion is to define the message type in the watcher (as it is now) and read it from there in the main where you attach the handlers

app/process/watcher/consensus-message/consensus-message.go
```diff
@@ -88,15 +90,29 @@ func (ctw ConsensusTopicWatcher) subscribeToTopic(q *queue.Queue) {
 	log.Infof("Started Consensus Message Watcher for topic [%s]\n", ctw.topicID)
 	unprocessedMessages, err := ctw.client.GetUnprocessedMessagesAfterTimestamp(ctw.topicID, milestoneTimestamp)
 	if err != nil {
+		log.Errorf("Could not get unprocessed messages after timestamp [%s]", milestoneTimestamp)
 		log.Fatal(err)
 	}
 	ctw.started = true
+	log.Println(unprocessedMessages)
```
Daniel-K-Ivanov, line 97: You should remove this log or add more concrete info. F.e `Found N unprocessed messages. Processing now` or something like that

app/process/watcher/consensus-message/consensus-message.go
```diff
@@ -88,15 +90,29 @@ func (ctw ConsensusTopicWatcher) subscribeToTopic(q *queue.Queue) {
...
 	for _, u := range unprocessedMessages.Messages {
-		// validate and check body
-		msg := &protomsg.TopicSignatureMessage{}
+		decodedMessage, err := b64.StdEncoding.DecodeString(u.Message)
+		if err != nil {
+			log.Errorf("Could not decode message - [%s]", u.Message)
```
Daniel-K-Ivanov, line 101: I think that we should add info that we are going to skip the processing of this message:
"Could not decode message - [%s]. Skipping the processing of this message", u.Message

app/process/watcher/consensus-message/consensus-message.go
```diff
@@ -88,15 +90,29 @@ func (ctw ConsensusTopicWatcher) subscribeToTopic(q *queue.Queue) {
...
+			continue
+		}
+
+		msg := &validatorproto.TopicSignatureMessage{}
+		err = proto.Unmarshal(decodedMessage, msg)
+		if err != nil {
+			log.Errorf("Could not unmarshal message - [%s] - [%s]", u.Message, err)
```
Daniel-K-Ivanov, line 108: same comment for skipping processing of the message

app/process/watcher/consensus-message/consensus-message.go
```diff
@@ -88,15 +90,29 @@ func (ctw ConsensusTopicWatcher) subscribeToTopic(q *queue.Queue) {
...
+			continue
+		}
+
 		publisher.Publish(msg, ctw.typeMessage, ctw.topicID, q)
-		err := ctw.statusRepository.UpdateLastFetchedTimestamp(ctw.topicID.String(), u.ConsensusTimestamp)
+		err = ctw.statusRepository.UpdateLastFetchedTimestamp(ctw.topicID.String(), u.ConsensusTimestamp)
 		if err != nil {
+			log.Errorf("Could not update last fetched timestamp - [%s]", u.ConsensusTimestamp)
 			log.Fatal(err)
```
Daniel-K-Ivanov, line 116: I do not think that we should do fatal here. The whole service will crash

app/process/watcher/consensus-message/consensus-message.go
```diff
@@ -107,10 +123,15 @@ func (ctw ConsensusTopicWatcher) subscribeToTopic(q *queue.Queue) {
...
+				if err != nil {
+					log.Fatal(err)
+				}
+
 				publisher.Publish(msg, ctw.typeMessage, ctw.topicID, q)
-				err := ctw.statusRepository.UpdateLastFetchedTimestamp(ctw.topicID.String(), strconv.FormatInt(response.ConsensusTimestamp.Unix(), 10))
+				err = ctw.statusRepository.UpdateLastFetchedTimestamp(ctw.topicID.String(), strconv.FormatInt(response.ConsensusTimestamp.Unix(), 10))
 				if err != nil {
 					log.Fatal(err)
 				}
```
Daniel-K-Ivanov, line 137: This logic seems very similar to the one you are using when you are handling unprocessed messages. Maybe you can extract it to a function?

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,110 @@
...
+	}
+
+	ctm := &validatorproto.CryptoTransferMessage{
+		TransactionId: m.TransactionId,
+		EthAddress:    m.EthAddress,
+		Amount:        m.Amount,
+		Fee:           m.Fee,
+	}
+
+	log.Printf("[%s] signature", m.GetSignature())
```
Daniel-K-Ivanov, line 51: If we are looking at the logs, we are going to see random strings of signatures.
I think that first of all we should not log the signatures, and second, if we logging anything we should add transaction ID to every log. Otherwise, we won't know what the context is. Think of how the logs will look like if there are 10 txns/sec.

Daniel-K-Ivanov, line 51: If we are looking at the logs we are going to miss the context. We should log something like:
`New Consensus Message for processing Transaction ID [ID]  was received`

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,110 @@
...
+
+	decodedSig, err := hex.DecodeString(m.GetSignature())
+	if err != nil {
+		return errors.New(fmt.Sprintf("Failed to decode signature. - [%s]", err))
+	}
+
+	hash := crypto.Keccak256([]byte(fmt.Sprintf("%s-%s-%d-%s", ctm.TransactionId, ctm.EthAddress, ctm.Amount, ctm.Fee)))
+
+	log.Printf("[%s] hex hash", hex.EncodeToString(hash))
```
Daniel-K-Ivanov, line 60: Same comment for the logs are the first one. I do not think that we should log the hashes.

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,110 @@
...
+		EthAddress:    m.EthAddress,
+		Amount:        m.Amount,
+		Fee:           m.Fee,
+	}
+
+	log.Printf("[%s] signature", m.GetSignature())
+
+	decodedSig, err := hex.DecodeString(m.GetSignature())
+	if err != nil {
+		return errors.New(fmt.Sprintf("Failed to decode signature. - [%s]", err))
```
Daniel-K-Ivanov, line 55: We need to add info about the TXID in the log, you can add the signature as-well

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,110 @@
...
+	}
+
+	hash := crypto.Keccak256([]byte(fmt.Sprintf("%s-%s-%d-%s", ctm.TransactionId, ctm.EthAddress, ctm.Amount, ctm.Fee)))
+
+	log.Printf("[%s] hex hash", hex.EncodeToString(hash))
+
+	key, err := crypto.Ecrecover(hash, decodedSig)
+	if err != nil {
+		return errors.New(fmt.Sprintf("Failed to recover public key. - [%s]", err))
```
Daniel-K-Ivanov, line 64: We need to add info about the TXID in the log you can add the hash as-well

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,110 @@
...
+	}
+
+	pubKey, err := crypto.UnmarshalPubkey(key)
+	if err != nil {
+		return errors.New(fmt.Sprintf("Failed to unmarshal public key. - [%s]", err))
```
Daniel-K-Ivanov, line 69: We need to add info about the TXID in the log you can add the key as-well

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,110 @@
...
+	}
+
+	address := crypto.PubkeyToAddress(*pubKey)
+
+	if !cmh.isValidAddress(address.String()) {
+		return errors.New(fmt.Sprintf("Address is not valid - [%s]", address.String()))
```
Daniel-K-Ivanov, line 75: We need to add info about the TXID in the log you can add the address as-well

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,110 @@
...
+		Amount:        m.Amount,
+		Fee:           m.Fee,
+		Signature:     m.Signature,
+		Hash:          hex.EncodeToString(hash),
+	})
+	if err != nil {
+		return errors.New(fmt.Sprintf("Could not add Transaction Message with Transaction Id and Signature - [%s]-[%s]", m.TransactionId, m.Signature))
+	}
+
+	fmt.Println("Success.")
```
Daniel-K-Ivanov, line 99: We must log the TX ID as-well.
Something like:
`Successfully verified and persisted TX with ID [ID]`

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,110 @@
...
+		return errors.New(fmt.Sprintf("Failed to unmarshal public key. - [%s]", err))
+	}
+
+	address := crypto.PubkeyToAddress(*pubKey)
+
+	if !cmh.isValidAddress(address.String()) {
+		return errors.New(fmt.Sprintf("Address is not valid - [%s]", address.String()))
+	}
+
+	messages, err := cmh.repository.GetByTxIdAndSignature(m.TransactionId, m.Signature)
```
Daniel-K-Ivanov, line 78: I think that for our next logic -> processing the TX if all signatures are acquired by the soft elected leader, we should get all TXns with this ID. In such case it would be better to change the `GetByTxIdAndSignature` to `GetTransaction` imo
and checking for duplicated signatures in the code here

Daniel-K-Ivanov (changes requested):

app/process/handler/consensus-message/consensus-message.go
georgiyazovaliiski: We only have the hash in bytes at this stage of the method. Should I encode and log it?

app/process/handler/consensus-message/consensus-message.go
georgiyazovaliiski: I will change the name of the method, but the duplication check is already done in the other branch for PR2.

app/process/watcher/consensus-message/consensus-message.go
```diff
@@ -81,6 +81,21 @@ func (ctw ConsensusTopicWatcher) getTimestamp(q *queue.Queue) string {
 	return milestoneTimestamp
 }
 
+func (ctw ConsensusTopicWatcher) updateTimestamp(message []byte, timestamp string, q *queue.Queue) {
```
Daniel-K-Ivanov, line 84: since we are not only updating the Timestamp, but actually publishing to the channel as-well (I would say the core of the logic), I think that a better name to the method would be `processMessage`

Daniel-K-Ivanov (approved):

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -1,3 +1,114 @@
...
+}
+
+func (cmh ConsensusMessageHandler) Recover(queue *queue.Queue) {
+	log.Println("Recovery method not implemented yet.")
+}
+
+func NewConsensusMessageHandler(r repositories.MessageRepository) *ConsensusMessageHandler {
+	return &ConsensusMessageHandler{
+		repository:     r,
+		validAddresses: config.LoadConfig().Hedera.Handler.ConsensusMessage.Addresses,
```
failfmi, line 30: It would be good to validate configuration addresses (regex checks).

failfmi (approved):

app/process/handler/consensus-message/consensus-message.go
georgiyazovaliiski: Can we do that in the next PR?

PR #40, opened 2020-11-30
Fix(eth-signer): signature decoding

failfmi:
**Detailed description**:
* Decrement recovery bit

Fixes #None

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/app/helper/ethereum/helper.go
+++ b/app/helper/ethereum/helper.go
@@ -1,6 +1,7 @@
 package ethereum
 
 import (
+	"encoding/hex"
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/limechain/hedera-eth-bridge-validator/app/helper"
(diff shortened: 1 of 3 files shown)


app/helper/ethereum/helper.go
```diff
@@ -60,3 +62,24 @@ func EncodeData(ctm *proto.CryptoTransferMessage) ([]byte, error) {
 		amountBn,
 		feeBn)
 }
+
+func DecodeSignature(signature string) (decodedSignature []byte, ethSignature []byte, err error) {
```
Daniel-K-Ivanov, line 66: If this one is `DecodeSignature` should the one above be `EncodeSignature` instead of `EncodeData`?

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -85,25 +85,25 @@ func (cmh ConsensusMessageHandler) handlePayload(payload []byte) error {
 		return errors.New(fmt.Sprintf("[%s] - Address is not valid - [%s]", m.TransactionId, address.String()))
 	}
 
-	messages, err := cmh.repository.GetTransaction(m.TransactionId, m.Signature)
+	messages, err := cmh.repository.GetTransaction(m.TransactionId, hex.EncodeToString(ethSig))
```
Daniel-K-Ivanov, line 88: maybe we can extract a variable `hexEncodedEthSig` from `hex.EncodeToString(ethSig)`. We are using it in 4-5 different places

app/helper/ethereum/helper.go
failfmi: Encode data is used to rebuild the hash which is signed.

PR #41, opened 2020-12-01
Feature: Eth signatures submission + Cancellation

failfmi:
**Detailed description**:
* Add scheduler eth submission
* Add scheduler hcs eth tx submission
* Add topic handler cancellation on equal txs
* Update protobufs
* Docker build: update image and install `ca-certificates`

Fixes #None

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


app/services/scheduler/scheduler.go
```diff
@@ -88,10 +131,67 @@ func (s *Scheduler) computeExecutionTime(messages []message.TransactionMessage)
...
+	if err != nil {
+		s.logger.Errorf("Waiting for execution for TX [%s] and Hash [%s] failed. Error [%s].", id, ethTxString, err)
+		return
+	}
+
+	if success {
+		s.logger.Infof("Successful execution of TX [%s] with TX Hash [%s].", id, ethTxString)
+	} else {
+		s.logger.Warn("Execution for TX [%s] with TX Hash [%s] was not successful.", id, ethTxString)
+	}
```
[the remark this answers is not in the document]
Daniel-K-Ivanov: and other thing related to the name of the function. 
we are not actually waiting for success but rather for the tx to be mined. Maybe we could rename the function to `waitForEthTxMined`, dunno. I don't have a strong opinion about it

Daniel-K-Ivanov (changes requested):

app/clients/ethereum/ethereum.go
```diff
@@ -58,6 +60,20 @@ func (ec *EthereumClient) waitForTransactionReceipt(hash common.Hash) (txReceipt
 	return ec.Client.TransactionReceipt(context.Background(), hash)
 }
 
+func (ec *EthereumClient) GetTransactionFromAddress(hash common.Hash) (string, error) {
```
Daniel-K-Ivanov, line 63: Isn't this actually `GetTransactionSender`?

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -87,7 +91,12 @@ func (cmh ConsensusMessageHandler) errorHandler(payload []byte) {
 func (cmh ConsensusMessageHandler) handleEthTxMessage(m *validatorproto.TopicEthTransactionMessage) error {
 	// TODO: verify authenticity of transaction hash
 
-	return cmh.scheduler.Cancel(m.TransactionId)
+	fromAddress, err := cmh.ethClient.GetTransactionFromAddress(common.HexToHash(m.EthTxHash))
```
Daniel-K-Ivanov, line 94: Will this return the TX even if it is still in the mempool?
What if the node is not synced fully and it returns that the tx is not present. Given the fast transactions in hedera, I think that this is plausible. Maybe we should consider setting the "from" in the proto file?

app/services/scheduler/scheduler.go
```diff
@@ -34,24 +34,27 @@ type Scheduler struct {
...
 	et, err := s.computeExecutionTime(submission.Messages)
 	if err != nil {
 		return err
 	}
 
 	executeIn := time.Until(et)
 	timer := time.NewTimer(executeIn)
-	s.tasks.Store(id, timer)
+
+	_, alreadyExisted := s.tasks.LoadOrStore(id, &scheduler.Storage{
```
Daniel-K-Ivanov, line 45: Do we need to store the submitter address if we have it in Scheduler.operator?

app/services/scheduler/scheduler.go
```diff
@@ -76,18 +89,20 @@ func (s *Scheduler) Schedule(id string, submission ethsubmission.Submission) err
...
 	if !exists {
 		s.logger.Warnf("Scheduled transaction execution for [%s] not found.", id)
 		return nil
 	}
-	s.tasks.Delete(id)
 
-	timer := t.(*time.Timer)
-	timer.Stop()
+	storage := t.(*scheduler.Storage)
+	storage.Timer.Stop()
```
Daniel-K-Ivanov, line 100: I think that stopping the timer should be inside the `if`. We should be stopping the timer only if the submitter is != current operator address

Daniel-K-Ivanov (approved):

georgiyazovaliiski (approved):

PR #65, opened 2021-02-03
Add -debug flag when running the main application

strenev:
Add -debug flag when running the main application. Default value is false, and if -debug is not set, the LogLevel is Info.


failfmi (approved):

failfmi:
LGTM! If you would like, you can create a `docs` folder and move the `run.md` there.

strenev:
> LGTM! If you would like, you can create a `docs` folder and move `run.md` there.

Yes, makes sense - moved it :)

georgiyazovaliiski:
LGTM

PR #94, opened 2021-02-19
Feature: ETH transaction authenticity

### LINKED ISSUE #79: Verify ETH TX Hash Authenticity
At the moment the `consensus-message` handler does not verify the TX Hash that is submitted to the topic. A validator can submit another TX and provide the Hash and this would cause the validators to unschedule all of their transactions.

Enhance the validator consensus-message handler by implementing a logic for verifying that the TX hash that is published in the HCS topic:
- Is targeting the right contract
- is calling the right function
- has the correct parameters

One way to do it might be to execute `estimateGas` function with the same parameters. Keep in mind that if we are going to do it this way, the `estimateGas` function might return revering TX because the original TX might get mined and the estimate gas would revert due to the checks in the mint function.

failfmi:
**Detailed description**:
* Add ETH tx authenticity

Fixes #79 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


app/process/handler/consensus-message/consensus-message.go
```diff
@@ -120,6 +133,65 @@ func (cmh ConsensusMessageHandler) handleEthTxMessage(m *validatorproto.TopicEth
...
+		return false, nil
+	}
+
+	dbTx, err := cmh.transactionRepository.GetByTransactionId(m.TransactionId)
+	if err != nil {
+		return false, err
+	}
+	if dbTx == nil {
+		return false, nil
+	}
```
georgiyazovaliiski, line 162: Can we merge these if statements?
In any way we are returning false as first return paratemer and whatever the error value is. (Might be null, might be a specific error)

georgiyazovaliiski (changes requested):

app/process/handler/consensus-message/consensus-message.go
failfmi: Well, not really. The first error returns the actual error, the second returns `false, nil`, which is not an actual error, just that it's actually `invalid`.

georgiyazovaliiski (dismissed):

app/helper/ethereum/helper.go
```diff
@@ -84,6 +105,64 @@ func DecodeSignature(signature string) (decodedSignature []byte, ethSignature st
...
+
+func DecodeBridgeMintFunction(data []byte) (transferMessage *proto.CryptoTransferMessage, signatures [][]byte, err error) {
+	bridgeAbi, err := abi.JSON(strings.NewReader(bridge.BridgeABI))
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// bytes transactionId, address receiver, uint256 amount, uint256 fee, bytes[] signatures
+	decodedParameters := make(map[string]interface{})
+	err = bridgeAbi.Methods[MintFunction].Inputs.UnpackIntoMap(decodedParameters, data[4:])
```
Daniel-K-Ivanov, line 119: nit: we could add a comment that [4:] strips the function signature

app/process/handler/consensus-message/consensus-message.go
```diff
@@ -120,6 +133,70 @@ func (cmh ConsensusMessageHandler) handleEthTxMessage(m *validatorproto.TopicEth
...
+
+	if strings.ToLower(tx.To().String()) != strings.ToLower(cmh.bridgeContractAddress) {
+		cmh.logger.Debugf("[%s] - ETH TX [%s] - Failed authenticity - Different To Address [%s].", m.TransactionId, m.EthTxHash, tx.To().String())
+		return false, nil
+	}
+
+	txMessage, signatures, err := ethhelper.DecodeBridgeMintFunction(tx.Data())
+	if err != nil {
+		return false, err
+	}
```
Daniel-K-Ivanov, line 151: `Decode` can return `ErrorInvalidMintFunctionParameters`. This would mean that here `false` and `ErrorInvalidMintFunctionParameters` will be returned as result, and in the upper logic you are logging error and returning if `err` is returned. 

I think that `ErrorInvalidMintFunctionParameters` must be interpreted as `Eth TX not authentic` and not an error.

Daniel-K-Ivanov (changes requested):

Daniel-K-Ivanov (approved):

Daniel-K-Ivanov (approved):

PR #97, opened 2021-02-22
Feature: expose metadata api call

### LINKED ISSUE #78: Expose REST endpoint MEMO info
**Problem:**
If the user does not provide correctly encoded memo the funds will be sent to the custodial account, but the transaction won't be processed. There are several ways a user can mess up the memo, so that the node rejects to process it. The most obvious one is providing insufficient fee or invalid gas price.

**Solution**
One solution for resolving the problem of sending funds and not get their transactions processed is by introducing an endpoint in the Validator Node.

Example:
Intorduce `/api/v1/metadata` endpoint. Once called it will return the following object:
```Json
{
    "txFee": "X",
    "txFeeCurrency": "HBAR",
    "gasPriceGwei": "Y"
}
```

where `txFee` is the "suggested" tx fee for the transaction at the moment and `gasPrice` is the suggested gas price at the moment (slow gas price from Etherscan f.e).
`txFee` can be calculated based on the newest `FeeCalculator` logic. At the moment, the fee calculator does the following:
By a given `txFee`, `amountToTransfer` and `gasPriceGwei` provided in the memo, verifies that:
`txFee > estimatedFee`, where `estimatedFee=((hardcoded_gas_usage_in_config)*(specified_gas_price)) * HBAR/ETH rate`.

for this API to work, we need another method to be implemented in the calculator that returns the `txFee` and `gasPrice` that should be used right now.
How to calculate:
- for `gasPriceGwei` get the current `slow` gas price from etherscan coingecko if they have it so that we have only one external 3rd party API integration
based on that `gasPriceGwei, multiply it to the `hardcoded_gas_usage_in_config`, this results in `ETH(wei) required for the TX`.
- get the current rate between `HBAR/ETH`, convert it to `tinyBar/wei`
- convert the `ETH Required for the TX` to `tinybars` required by the TX using the `tinyBar/wei` currency rate
- as a result, return the calculated result (`tinybars required by the TX`) + the `gasPriceGwei` used for the calculation.

failfmi:
**Detailed description**:
* Expose REST endpoint to calculate suggested tx fee based on hbar/eth rate + validators gas usage

Closes #78 

**Checklist**
- [ ] Documentation added
- [x] Tests updated


app/router/router.go
```diff
@@ -0,0 +1,43 @@
...
+		c.Handler)
+
+	return &APIRouter{
+		Router: router,
+	}
+}
+
+func (api *APIRouter) AddV1Router(router http.Handler) {
+	api.Router.Mount(apiV1, router)
+}
```
Daniel-K-Ivanov, line 43: as noted in the other comment, I think that the router should mount the endpoints, or at least have a point in which the `main` calls only `initialise` on the API and 1) router is created 2) handlers are mounted

Daniel-K-Ivanov (changes requested):
Now that we have the endpoint, we could use it in the e2e tests to figure out the `hbar` tx fee. This would make the e2e tests more stable. Not failing due to hbar/eth price or gas price changes

e2e/e2e_test.go
```diff
@@ -57,23 +70,101 @@ func Test_E2E(t *testing.T) {
...
+	}
+
+	transactionResponse, whbarBalanceBefore := verifyCryptoTransfer(memo, acc, receiving, whbarInstance, whbarReceiverAddress, client, t)
+
+	ethTransactionHash := verifyTopicMessages(topicID, client, transactionResponse, expectedValidatorsCount, t)
+
+	verifyEthereumTXExecution(ethTransactionHash, whbarInstance, whbarReceiverAddress, whbarAmount.Int64(), whbarBalanceBefore, ethClient, t)
+}
+
+func calculateWHBarAmount(txFee string, percentage *big.Int) (*big.Int, error) {
```
Daniel-K-Ivanov, line 111: nit: rename this function to `calculateExpectedWHBarAmount`

Daniel-K-Ivanov (approved):

PR #103, opened 2021-02-24
Auto update config

### LINKED ISSUE #86: Automatically Update Contract-level Config
At the moment, there are several properties that are controlled on the contract level, however, they are set in the configuration `application.yaml` file (hardcoded). Those properties include:
- `hedera.client.operator.service_fee_percent` -> this can be read from the Smart Contract
- `hedera.handler.consensusmessage.addresses` -> the currently set validators in the contract. This can be read from the contract as-well

Issue:
If the validator set is changed in the contract, the node must be aware of that so that the number of signatures threshold is properly 
 calculated, the integrity of the auth signatures is validated correctly and the fee calculation is property estimated.

Tasks:
- add new function to the `ethereum bridge service`: `loadConfig`. This function must be called on node start and must:
1) Query the `serviceFee` from the contract; 2) Set the `service fee` in-memory in the `EthereumBridgeService`
1) Query the `custodians` from the contract; 2) Set the `custodians` in-memory in the `EthereumBridgeService` 
- add new functions to the current `ethereum bridge service`: `WatchFeeChangedEvent` and `WatchCustodianChanged`
- When the `ethereum watcher` is started ("listenForEvents") we must subscribe to the `WatchFeeChangedEvent` and `WatchCustodianChanged` events as-well (not only for the burn event). Once triggered, the appropriate property must be updated (f.e service fee / custodian) in the `EthereumBridgeService`. Note: For the update of the properties, use Mutex and log the update
- Update the fee-calculator in such away that it calls the `EthereumBridgeService` to read the `serviceFee` property and the `custodians` count 
- Update the CryptoTransferHandler to use the `EthereumBridgeService` for getting the `custodians` count
- Remove the `hedera.client.operator.service_fee_percent` and `hedera.handler.consensusmessage.addresses` from the `application.yml`

Holymir:
**Detailed description**:

- read serviceFee from smart contract
- read custodians from smart contract
- listen for ServiceFeeSet event and update serviseFee
- listen for MemberUpdated event and update custodians list

Fixes #86

**Checklist**
- [ ] Documentation added
- [x] Tests updated


app/process/handler/crypto-transfer/crypto-transfer_test.go
```diff
@@ -47,9 +47,19 @@ const (
...
+		"0xsomeaddress3",
+	}
+	// Value of the serviceFeePercent in percentage. Range 0% to 99.999% multiplied my 1000
+	serviceFeePercent uint64 = 10000
+)
+
 func getHederaConfig() config.Hedera {
 	hederaConfig := config.Hedera{}
-	hederaConfig.Client.ServiceFeePercent = 10
+	// hederaConfig.Client.ServiceFeePercent = 10
```
failfmi, line 62: This line is not needed anymore

failfmi (dismissed):

failfmi (dismissed):

app/services/ethereum/bridge/service.go
```diff
@@ -48,10 +54,18 @@ func NewBridgeContractService(client *ethclient.EthereumClient, config config.Et
...
+	bridge := &BridgeContractService{
 		Client:           client,
 		contractInstance: contractInstance,
 	}
+
+	bridge.UpdateMembers(nil)
+	bridge.UpdateServiceFee(nil)
+
+	go bridge.listenForMemberUpdatedEvent(nil)
+	go bridge.listenForChangeFeeEvent(nil)
```
Daniel-K-Ivanov, line 66: Lets update the functions so that they don't receive `opts *bind.Callops`. We can pass `nil` inside the functions directly to the underlying `bsc.contractInstance` f.e:
`bsc.contractInstance.ServiceFee(nil)`
`bsc.contractInstance.MembersCount(nil)`

Daniel-K-Ivanov, line 66: Lets update the functions so that they don't receive `opts *bind.Callops`. We can pass `nil` inside the functions directly to the underlying `bsc.contractInstance` f.e:
`bsc.contractInstance.ServiceFee(nil)`
`bsc.contractInstance.MembersCount(nil)`

app/services/ethereum/bridge/service.go
```diff
@@ -77,7 +91,76 @@ func (bsc *BridgeContractService) SubmitSignatures(opts *bind.TransactOpts, ctm
...
 
+func (bsc *BridgeContractService) GetCustodians() []string {
+	return bsc.custodians.Get()
+}
+
+func (bsc *BridgeContractService) GetServiceFee() *big.Int {
+	return bsc.servicefee.Get()
+}
+
+func (bsc *BridgeContractService) UpdateServiceFee(opts *bind.CallOpts) {
```
Daniel-K-Ivanov, line 102: I think that his should be private

app/services/ethereum/bridge/service.go
```diff
@@ -77,7 +91,76 @@ func (bsc *BridgeContractService) SubmitSignatures(opts *bind.TransactOpts, ctm
...
+	newFee, err := bsc.contractInstance.ServiceFee(opts)
+	if err != nil {
+		log.Error("Failed to get service fee.", err)
+	}
+	bsc.servicefee.Set(*newFee)
+}
+
+func (bsc *BridgeContractService) UpdateMembers(opts *bind.CallOpts) {
```
Daniel-K-Ivanov, line 110: I think that his should be private

app/services/ethereum/bridge/service.go
```diff
@@ -77,7 +91,76 @@ func (bsc *BridgeContractService) SubmitSignatures(opts *bind.TransactOpts, ctm
...
+}
+
+func (bsc *BridgeContractService) GetServiceFee() *big.Int {
+	return bsc.servicefee.Get()
+}
+
+func (bsc *BridgeContractService) UpdateServiceFee(opts *bind.CallOpts) {
+	newFee, err := bsc.contractInstance.ServiceFee(opts)
+	if err != nil {
+		log.Error("Failed to get service fee.", err)
```
Daniel-K-Ivanov, line 105: I think that if this fails, we will set invalid value to `serviceFee`. Maybe we should do `log.Fatal` since the initialisation of the contract variables will not be performed and we won't be able to run the validator node.

app/services/ethereum/bridge/service.go
```diff
@@ -77,7 +91,76 @@ func (bsc *BridgeContractService) SubmitSignatures(opts *bind.TransactOpts, ctm
...
+	}
+	bsc.servicefee.Set(*newFee)
+}
+
+func (bsc *BridgeContractService) UpdateMembers(opts *bind.CallOpts) {
+	membersCount, err := bsc.contractInstance.MembersCount(opts)
+	if err != nil {
+		log.Error("Failed to get members count.", err)
```
Daniel-K-Ivanov, line 113: Should be `log.Fatal`

app/services/ethereum/bridge/service.go
```diff
@@ -35,6 +39,8 @@ type BridgeContractService struct {
 	contractInstance *bridge.Bridge
 	Client           *ethclient.EthereumClient
 	mutex            sync.Mutex
+	custodians       custodians.Custodians
+	servicefee       servicefee.Servicefee
```
Daniel-K-Ivanov, line 43: You should include the `logger` here as property similar to Consensus Message Handler f.e
We want to have a context for the logger. That's why we do getLoggerFor in other files. 
```go
config.GetLoggerFor("Bridge Contract Service")
```

app/services/ethereum/bridge/service.go
```diff
@@ -77,7 +91,76 @@ func (bsc *BridgeContractService) SubmitSignatures(opts *bind.TransactOpts, ctm
...
+	membersCount, err := bsc.contractInstance.MembersCount(opts)
+	if err != nil {
+		log.Error("Failed to get members count.", err)
+	}
+
+	var newCustodiansArray []string
+	for i := 0; i < int(membersCount.Int64()); i++ {
+		addr, err := bsc.contractInstance.MemberAt(opts, big.NewInt(int64(i)))
+		if err != nil {
+			log.Error("Failed to get member address.", err)
```
Daniel-K-Ivanov, line 120: Should be log.Fatal

app/services/ethereum/bridge/service.go
```diff
@@ -77,7 +91,76 @@ func (bsc *BridgeContractService) SubmitSignatures(opts *bind.TransactOpts, ctm
...
+		}
+		newCustodiansArray = append(newCustodiansArray, addr.String())
+	}
+	bsc.custodians.Set(newCustodiansArray)
```
Daniel-K-Ivanov, line 124: Let's add log here that the members were updated

app/services/ethereum/bridge/service.go
```diff
@@ -77,7 +91,76 @@ func (bsc *BridgeContractService) SubmitSignatures(opts *bind.TransactOpts, ctm
...
+func (bsc *BridgeContractService) GetServiceFee() *big.Int {
+	return bsc.servicefee.Get()
+}
+
+func (bsc *BridgeContractService) UpdateServiceFee(opts *bind.CallOpts) {
+	newFee, err := bsc.contractInstance.ServiceFee(opts)
+	if err != nil {
+		log.Error("Failed to get service fee.", err)
+	}
+	bsc.servicefee.Set(*newFee)
```
Daniel-K-Ivanov, line 107: Let's add log here that the service fee is updated

app/services/ethereum/bridge/service.go
```diff
@@ -77,7 +91,76 @@ func (bsc *BridgeContractService) SubmitSignatures(opts *bind.TransactOpts, ctm
...
 func (bsc *BridgeContractService) WatchBurnEventLogs(opts *bind.WatchOpts, sink chan<- *bridge.BridgeBurn) (event.Subscription, error) {
 	var addresses []common.Address
 	return bsc.contractInstance.WatchBurn(opts, sink, addresses, [][]byte{})
 }
+
+func (bsc *BridgeContractService) listenForChangeFeeEvent(opts *bind.WatchOpts) {
+	events := make(chan *bridgecontract.BridgeServiceFeeSet)
+	sub, err := bsc.contractInstance.WatchServiceFeeSet(opts, events)
+	if err != nil {
+		log.Errorf("Failed to subscribe for WatchServiceFeeSet Event Logs for contract. Error [%s].", err)
```
Daniel-K-Ivanov, line 136: Should be `log.Fatal`

app/services/ethereum/bridge/service.go
```diff
@@ -77,7 +91,76 @@ func (bsc *BridgeContractService) SubmitSignatures(opts *bind.TransactOpts, ctm
...
+	}
+
+	for {
+		select {
+		case err := <-sub.Err():
+			log.Errorf("ServiceFeeSet Event Logs subscription failed. Error [%s].", err)
+			return
+		case eventLog := <-events:
+			bsc.servicefee.Set(*eventLog.NewServiceFee)
```
Daniel-K-Ivanov, line 145: We should add log here

app/services/ethereum/bridge/service.go
```diff
@@ -77,7 +91,76 @@ func (bsc *BridgeContractService) SubmitSignatures(opts *bind.TransactOpts, ctm
...
+		}
+	}
+}
+
+func (bsc *BridgeContractService) listenForMemberUpdatedEvent(opts *bind.WatchOpts) {
+	events := make(chan *bridgecontract.BridgeMemberUpdated)
+	sub, err := bsc.contractInstance.WatchMemberUpdated(opts, events)
+	if err != nil {
+		log.Errorf("Failed to subscribe for WatchMemberUpdated Event Logs for contract. Error [%s].", err)
```
Daniel-K-Ivanov, line 154: Should be log.fatal

Daniel-K-Ivanov (changes requested):

app/process/model/bridge/servicefee/servicefee.go
```diff
@@ -0,0 +1,39 @@
...
+
+package servicefee
+
+import (
+	"math/big"
+	"sync"
+)
+
+type Servicefee struct {
+	serviceFee big.Int
```
Daniel-K-Ivanov, line 25: I think that it would be better if we store the service fee in `float` and set the actually service fee in percentage. F.e the smart contract has 5000 set as a service fee and PRECISION `100000`. That would mean that we will store:
`5000/100000=0.05` here as variable. 

Reasoning:
The fee calculator should not know the PRECISION of the smart contract, even worse -> have it hardcoded in its code. Ideally the service fee should store the exact fee percentage

Daniel-K-Ivanov (changes requested):

failfmi (dismissed):

app/process/model/bridge/servicefee/servicefee.go
Daniel-K-Ivanov: Will be fixed in another PR

Daniel-K-Ivanov (dismissed):

Daniel-K-Ivanov (approved):

PR #108, opened 2021-03-01
Feature: Schedule transactions for Bridge Burn events

### LINKED ISSUE #98: Ethereum to Hedera bridging
1. Implement new `process/handler/burn-event/burn-event.go`
The service must be called from the `watcher/ethereum/ethereum.go` watcher

Once called, the `burn-event` handler should create new `CryptoTransfer` as `ScheduledTransaction`:
```
HbarSender: {account={custodialAccount}; amount={amount from burn event}}
HbarReceiver: {account={account from burn event}; amount={amount from burn event}}
nonce={ethereumTxHash}{eventLogIndex}
```
Set the `CryptoTransfer` as `transactionBody` of the `ScheduledTransaction` and submit it.

failfmi:
**Detailed description**:
* Update `hedera-sdk-go` version to `v2.1.5-beta.3`
* Add Ethereum watcher to push burn events to handler
* Add Burn event handler
* Fix configuration from `main`
* Fix e2e token test from `main`

Fixes #98 

**Checklist**
- [x] Documentation added
- [x] Tests updated


config/application.yml
```diff
@@ -11,6 +11,7 @@ hedera:
   eth:
     node_url:
     bridge_contract_address:
+    whbar_contract_address:
```
Daniel-K-Ivanov, line 14: Where are you using the whbar address?

config/config.go
```diff
@@ -77,8 +77,9 @@ type Hedera struct {
 }
 
 type Handler struct {
```
Daniel-K-Ivanov, line 79: Wouldn't it be better to name the handler `BurnEvent`? The other handlers are `CryptoTransfer` and `ConsensusMessage` -> they represent the event which we are handling

config/application.yml
```diff
@@ -41,3 +42,6 @@ hedera:
     consensus-message:
       topic_id:
       send_deadline: 300
+    scheduled-transaction:
```
Daniel-K-Ivanov, line 45: as stated in the other comment, it might be better to name the config as `burn-event`

failfmi:
Closing until it is ready for review

app/helper/helper.go
```diff
@@ -31,3 +33,21 @@ func ToBigInt(value string) (*big.Int, error) {
...
 	return amount, nil
 }
+
+// ToMirrorNodeTransactionID parses TX with format `0.0.X@{seconds}.{nanos}` to format `0.0.X-{seconds}-{nanos}`
+func ToMirrorNodeTransactionID(txId string) string {
+	split := strings.Split(txId, "?")
+	split = strings.Split(split[0], "@")
+	accId := split[0]
+	split = strings.Split(split[1], ".")
+	return fmt.Sprintf("%s-%s-%s", accId, split[0], fmt.Sprintf("%09s", split[1]))
```
georgiyazovaliiski, line 43: `split[0]` -> `fmt.Sprintf("%09s", split[0])`
The same issue happens with the seconds sometimes as well.

app/process/handler/burn-event/handler.go
```diff
@@ -0,0 +1,46 @@
...
+	return &Handler{
+		service: service,
+		logger:  config.GetLoggerFor("Scheduled Transaction Handler"),
+	}
+}
+
+func (sth Handler) Handle(payload interface{}) {
+	burnEvent, ok := payload.(*burn_event.BurnEvent)
+	if !ok {
+		sth.logger.Errorf("Error could not cast BurnEvent payload [%s]", payload)
```
georgiyazovaliiski, line 41: Errorf already displays there is an error. 
`sth.logger.Errorf("Error could not cast BurnEvent payload [%s]", payload)` -> `sth.logger.Errorf("Could not cast BurnEvent payload [%s]", payload)`

examples/three-validators/alice/config/application.yml
```diff
@@ -1,45 +1,11 @@
...
+# Check docs/configuration.md for more information
+#hedera:
+#  validator:
+#    db:
+#      host:
+#      name:
+#      password:
+#      port:
+#      username:
+#    port:
```
georgiyazovaliiski, line 11: Why do we need this?

examples/three-validators/bob/config/application.yml
```diff
@@ -1,45 +1,11 @@
...
+# Check docs/configuration.md for more information
+#hedera:
+#  validator:
+#    db:
+#      host:
+#      name:
+#      password:
+#      port:
+#      username:
+#    port:
```
georgiyazovaliiski, line 11: Why do we need this?

examples/three-validators/carol/config/application.yml
```diff
@@ -1,45 +1,11 @@
...
+# Check docs/configuration.md for more information
+#hedera:
+#  validator:
+#    db:
+#      host:
+#      name:
+#      password:
+#      port:
+#      username:
+#    port:
```
georgiyazovaliiski, line 11: Why do we need this?

georgiyazovaliiski (changes requested):

app/helper/helper.go
failfmi: If you knew what is the amount behind split[0], you would know it is not possible to be less than 10 symbols

examples/three-validators/alice/config/application.yml
failfmi: So you do not have to maintain 4 files

app/services/burn-event/service.go
```diff
@@ -0,0 +1,178 @@
...
+		return
+	}
+
+	receipt, err := txResponse.GetReceipt(s.hederaNodeClient.GetClient())
+	if err != nil {
+		s.logger.Errorf("[%s] - Failed to get transaction receipt for schedule sign [%s]. Error: [%s].", txHash, scheduleID, err)
+		return
+	}
+
+	if receipt.Status != hedera.StatusSuccess {
```
Daniel-K-Ivanov, line 145: I think that we should explicitly check the status of the receipt against `already executed`. 
If already executed -> log specific message with debug/trace that this TX ended up one of the last ones and the tx already executed.
If the status is != success and `alredy execute`, then we should log an error similar to what you are logging.

app/services/burn-event/service.go
```diff
@@ -0,0 +1,178 @@
...
+		s.logger.Errorf("[%s] - Failed to get transaction receipt for [%s]", event.TxHash, transactionResponse.TransactionID)
+		return
+	}
+
+	switch txReceipt.Status {
+	case hedera.StatusIdenticalScheduleAlreadyCreated:
+		s.handleScheduleSign(event.TxHash, helper.ToMirrorNodeTransactionID(txReceipt.ScheduledTransactionID.String()), *txReceipt.ScheduleID)
+	case hedera.StatusSuccess:
+		transactionID := helper.ToMirrorNodeTransactionID(txReceipt.ScheduledTransactionID.String())
+		s.logger.Infof("[%s] - Updating db status to Submitted with TransactionID [%s].", event.TxHash, transactionID)
```
Daniel-K-Ivanov, line 105: I think that it is better to log the update after it has occurred. That way we will keep it consistent as-well

config/application.yml
```diff
@@ -26,6 +26,8 @@ hedera:
       eth_private_key:
     base_gas_usage: 130000
     gas_per_validator: 54000
+    threshold_account:
```
Daniel-K-Ivanov, line 29: q: threshold_account is the account that we are watching in the transfer watcher right?

app/helper/helper.go
```diff
@@ -31,3 +33,21 @@ func ToBigInt(value string) (*big.Int, error) {
 
 	return amount, nil
 }
+
+// ToMirrorNodeTransactionID parses TX with format `0.0.X@{seconds}.{nanos}` to format `0.0.X-{seconds}-{nanos}`
+func ToMirrorNodeTransactionID(txId string) string {
```
Daniel-K-Ivanov, line 38: - what do you think about the following approach when it comes to helper's:
https://play.golang.org/p/Ey07k5HGSGV
I think that helper is becoming the go to file when we do not see a proper place to fit the functionality and this would evolve into a code that it is hard to maintain. 

- similar function is placed in the `mirror node` client. Maybe we could use the pattern shown above and extract them into something coherent -> `TransactionId` or whatelse in a separate package

Daniel-K-Ivanov (changes requested):

app/persistence/entity/burn-event/status.go
```diff
@@ -17,10 +17,14 @@
...
 	// This is a terminal status
 	StatusCompleted = "COMPLETED"
-	StatusFailed    = "FAILED"
-	StatusInitial   = "INITIAL"
+	// StatusFailed is a status set once the BurnEvent operation has failed
+	// This is a terminal status
+	StatusFailed = "FAILED"
+	// StatusInitial is the initial status on BurnEvent Record creation
+	StatusInitial = "INITIAL"
+	// StatusSubmitted is set once the Hedera Scheduled Transaction (Create/Sign) is submitted to HCS
```
Daniel-K-Ivanov, line 28: nit: we are not submitting to `HCS`

app/services/burn-event/service.go
```diff
@@ -122,37 +126,40 @@ func (s Service) ProcessEvent(event burn_event.BurnEvent) {
...
 
-	transactionID := helper.ToMirrorNodeTransactionID(txReceipt.ScheduledTransactionID.String())
+	transactionID := hederahelper.ToMirrorNodeTransactionID(txReceipt.ScheduledTransactionID.String())
+	err = s.repository.UpdateStatusSubmitted(event.Id, txReceipt.ScheduleID.String(), transactionID)
+	if err != nil {
+		s.logger.Errorf(
+			"[%s] - Failed to update submitted status with TransactionID [%s], ScheduleID [%s]. Error [%s].",
+			event.Id, transactionID, txReceipt.ScheduleID.String(), err)
+		return
+	}
```
Daniel-K-Ivanov, line 136: We are updating `status submitted` twice for `hedera.StatusSuccess`

Daniel-K-Ivanov (changes requested):

Daniel-K-Ivanov (dismissed):

Daniel-K-Ivanov (approved):

PR #129, opened 2021-03-11
Generalized Bridge - Token Support

### LINKED ISSUE #125: Generalised Bridge - Token Support
Update the Validator node in such a way that:
- Crypto Transfer Watcher is refactored to Transfer Watcher - watching not only hbar transfers but token transfers as-well
- Crypto Transfer Handler is refactored to Transfer handler - handling not only hbar transfers but token transfers as-well
- Add new column in the Transaction record in DB for specifying the ERC20 Contract address
- Update the signed message to include the token address and publish it to HCS
- Update the Consensus Message handler to parse the new signature format and be able to support transfer messages for tokens as-well.
- Add new column in HCS Messages records in DB for populating storing the ERC20 contract address

georgiyazovaliiski:
**Detailed description**:
This PR modifies the already implemented logic and extends it to support "listening" to token transfers.

**Not Tested!**
This implementation is still not tested, because a mirror-node api problem is still being resolved.

Fixes #125 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


app/process/watcher/transfer/transfer.go
```diff
@@ -154,17 +150,35 @@ func (ctw CryptoTransferWatcher) beginWatching(q *queue.Queue) {
...
+			asset = tokenTr.Asset
 			break
 		}
 	}
 
+	if asset == "" {
+		for _, tr := range tx.Transfers {
+			if tr.Account == ctw.accountID.String() {
+				amount = fmt.Sprint(tr.Amount)
+				asset = "HBAR"
```
failfmi, line 176: And this to be `ctw.contractService.getWHBARAddress`

app/persistence/transaction/transaction.go
```diff
@@ -44,6 +44,7 @@ type Transaction struct {
 	SubmissionTxId string
 	Status         string
 	EthHash        string
+	Asset          string
```
Daniel-K-Ivanov, line 47: I am wondering whether it will be better to populate `asset` with `HBAR` (for Hbars) and `0.0.15` for tokens OR populate the `erc20-contract-address`.

@georgiyazovaliiski @failfmi wdut?

app/process/handler/transfer/transfer.go
```diff
@@ -70,11 +66,12 @@ func NewHandler(
 		transactionRepo:    transactionRepository,
 		logger:             config.GetLoggerFor("Account Transfer Handler"),
 		feeCalculator:      feeCalculator,
+		contractService:    contractService,
 	}
 }
 
 // Recover mechanism
-func (cth Handler) Recover(q *queue.Queue) {
+func (cth *Handler) Recover(q *queue.Queue) {
```
Daniel-K-Ivanov, line 74: nit: no need for `*`. We do not set anything on the `Handler` struct so we do not need the `*`

app/process/model/transaction/transaction.go
```diff
@@ -18,22 +18,24 @@ package transaction
...
+		ValidDurationSeconds string     `json:"valid_duration_seconds"`
+		Node                 string     `json:"node"`
+		TransactionID        string     `json:"transaction_id"`
+		Transfers            []Transfer `json:"transfers"`
+		TokenTransfers       []Transfer `json:"token_transfers"`
 	}
 	Transfer struct {
 		Account string `json:"account"`
 		Amount  int64  `json:"amount"`
+		Asset   string `json:"token_id"`
```
Daniel-K-Ivanov, line 38: Maybe we should add comment here that for `Transfers` this property is not populated

proto/topic_eth_signature_message.proto
```diff
@@ -10,4 +10,5 @@ message TopicEthSignatureMessage {
   string amount = 3;
   string fee = 4;
   string signature = 5;
+  string asset = 6;
```
Daniel-K-Ivanov, line 13: I think that here we should specify that it is `erc20Address`, because that is the data that will be signed in the signature

Daniel-K-Ivanov (changes requested):

app/persistence/transaction/transaction.go
failfmi: If you would like to have information for both the asset on Hedera(either `HBAR` or `TokenID`) and the corresponding `ERC20`, we can introduce two properties.

Daniel-K-Ivanov: having both might be redundant... I would go for `HBAR` / `0.0.15` compared to `ERC20-contract-address`. 
+ we will have `erc20-address` as column in the `messages` records (since it will be signed)

PR #148, opened 2021-03-17
Gas price update

### LINKED ISSUE #135: Submit the signatures with the specified Gas Price from memo
At the moment, when we are submitting validator signatures to the bridge contract (`SubmitSignatures(opts *bind.TransactOpts, txId, ethAddress, amount, fee string, signatures [][]byte) (*types.Transaction, error)`) we are not submitting them at the `user` specified gas price from the memo, but using the default mechanism for figuring out gas price. We should change that

radtonev:
**Detailed description**:
Submit signatures with gas price provided by memo.

Fixes #135

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


app/services/messages/service.go
```diff
@@ -194,6 +194,24 @@ func (ss *Service) prepareEthereumMintTask(txId string, ethAddress string, amoun
...
+			gasPriceGwei, isSuccessful := new(big.Int).SetString(t.GasPriceGwei, 10)
+			if !isSuccessful {
+				ss.logger.Errorf("Failed to parse provided gas price for TX [%s]. Error: %s", txId, err)
+				return
+			}
+			//Convert GWei to Wei
+			mul := new(big.Int).SetUint64(1000000000)
+			ethTransactor.GasPrice = new(big.Int).Mul(gasPriceGwei, mul)
+		}
+
```
failfmi, line 214: I do not think it is necessary to get the transaction again. Passing the `gasPrice` as input parameter and then call `ethTransactor.GasPrice = gasPrice` would to the trick.
In `ScheduleEthereumTxForSubmission` `GasPriceGwei` will be retrieved and converted to `wei`.
Let's wait [this](https://github.com/LimeChain/hedera-eth-bridge-validator/issues/139) to be resolved, and then update it.

app/services/messages/service.go
```diff
@@ -194,6 +194,24 @@ func (ss *Service) prepareEthereumMintTask(txId string, ethAddress string, amoun
 			ss.logger.Errorf("Failed to establish key transactor. Error %s", err)
 			return
 		}
+
+		if ethTransactor.GasPrice == nil {
+			//Set gas price from memo
+			t, err := ss.transactionRepository.GetByTransactionId(txId)
```
Daniel-K-Ivanov, line 200: Add TODO comment here to be addressed by @georgiyazovaliiski task

Daniel-K-Ivanov (approved):

PR #150, opened 2021-03-18
(#139) GasPrice Column in Message Record and Refactored Recovery Process

### LINKED ISSUE #139: Add gasprice column in the messages record
- Add `gasPrice` column in the messages record
-  (Once contracts are updated) Sign the gas price as-well
- Refactor the recovery process `processUnfinishedOperations` to query the messages repo and perform the check on whether the signature should be submitted or not

georgiyazovaliiski:
**Detailed description**:
I added a GasPrice Column in the Message Record table. GasPrice is now removed from the Transaction Records. Also, the recovery process is now refactored, according to the new structure of the project

Fixes #139 

Recovery Process is still not extensively tested.

**Checklist**
- [ ] Documentation added
- [x] Tests updated


app/encoding/auth-message/auth-message.go
```diff
@@ -38,9 +38,13 @@ func EncodeBytesFrom(txId, receiverEthAddress, erc20Address, amount, fee string)
...
 		return nil, err
 	}
+	gasPriceBn, err := helper.ToBigInt(gasPriceWei)
+	if err != nil {
+		return nil, err
+	}
 
 	// TODO: add common.HexToAddress(erc20Address) after contracts are updated
-	bytesToHash, err := args.Pack([]byte(txId), common.HexToAddress(receiverEthAddress), amountBn, feeBn)
+	bytesToHash, err := args.Pack([]byte(txId), common.HexToAddress(receiverEthAddress), amountBn, feeBn, gasPriceBn)
```
failfmi, line 47: Additional TODO in order e2e tests to pass

app/services/messages/service.go
```diff
@@ -195,26 +197,33 @@ func (ss *Service) prepareEthereumMintTask(txId string, ethAddress string, amoun
...
+		//	}
+		//	gasPriceGwei, isSuccessful := new(big.Int).SetString(t.GasPriceGwei, 10)
+		//	if !isSuccessful {
+		//		ss.logger.Errorf("Failed to parse provided gas price for TX [%s]. Error: %s", txId, err)
+		//		return
+		//	}
+		//	//Convert GWei to Wei
+		//	mul := new(big.Int).SetUint64(1000000000)
+		//	ethTransactor.GasPrice = new(big.Int).Mul(gasPriceGwei, mul)
+		//}
```
failfmi, line 226: This is no longer needed

app/services/messages/service.go
```diff
@@ -195,26 +197,33 @@ func (ss *Service) prepareEthereumMintTask(txId string, ethAddress string, amoun
 			return
 		}
 
+		// TODO: Refactored according to the comment below
```
failfmi, line 200: `TODO` is no longer needed.

app/services/messages/service.go
```diff
@@ -366,7 +375,7 @@ func (ss *Service) VerifyEthereumTxAuthenticity(tm encoding.TopicMessage) (bool,
 		return false, nil
 	}
 	// Verify Ethereum TX `call data`
-	txId, ethAddress, amount, fee, signatures, err := ethhelper.DecodeBridgeMintFunction(tx.Data())
+	txId, ethAddress, amount, fee, gasPrice, signatures, err := ethhelper.DecodeBridgeMintFunction(tx.Data())
```
failfmi, line 378: This will not return `gasPrice`.  Retrieve it from `tx.GasPrice()`

app/persistence/transaction/transaction.go
```diff
@@ -82,7 +81,6 @@ type Transaction struct {
 	EthTxMsgStatus        string
 	EthTxStatus           string
 	EthHash               string
-	GasPriceGwei          string
```
failfmi, line 85: This should be converted to `GasPriceWei`

app/services/messages/service.go
failfmi: L401: Add check to `if`: `dbTx.GasPriceWei != tx.GasPrice()`.

app/helper/ethereum/helper.go
```diff
@@ -55,21 +55,21 @@ func DecodeSignature(signature string) (decodedSignature []byte, ethSignature st
 	return switchSignatureValueV(decodedSig)
 }
 
-func DecodeBridgeMintFunction(data []byte) (txId, ethAddress, amount, fee string, signatures [][]byte, err error) {
+func DecodeBridgeMintFunction(data []byte) (txId, ethAddress, amount, fee, gasPrice string, signatures [][]byte, err error) {
```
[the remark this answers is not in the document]
failfmi: `erc20Address` will be retrieved from here though.

failfmi:
`SanityCheckSignature` will need to check `GasPriceWei` and `Erc20Address`

app/services/transfers/service.go
```diff
@@ -126,106 +128,132 @@ func (bs *Service) SaveRecoveredTxn(txId, amount string, asset string, m memo.Me
...
-		bs.logger.Errorf("Failed to encode the authorisation signature for TX ID [%s]. Error: %s", tm.TransactionId, err)
 		return err
 	}
-	signatureBytes, err := bs.ethSigner.Sign(authMsgHash)
+	gasPriceWei := ethereum.GweiToWei(gasPriceGWeiBn).String()
+
+	signature, err := ts.SignAuthorizationMessage(tm.TransactionId, tm.EthAddress, tm.Erc20Address, tm.Amount, tm.Fee, gasPriceWei)
 	if err != nil {
-		bs.logger.Errorf("Failed to sign the authorisation signature for TX ID [%s]. Error: %s", tm.TransactionId, err)
+		ts.logger.Errorf("Failed to Validate and Sign TransactionID [%s]. Error [%s].", tm.TransactionId, err)
```
failfmi, line 167: Validate?

failfmi (approved):

app/services/messages/service.go
```diff
@@ -90,17 +90,26 @@ func (ss *Service) SanityCheckSignature(tm encoding.TopicMessage) (bool, error)
...
+		ss.logger.Errorf("Provided Asset is not supported - [%s]", t.Asset)
+		return false, err
+	}
+
 	match := t.EthAddress == topicMessage.EthAddress &&
 		t.Amount == topicMessage.Amount &&
-		t.Fee == topicMessage.Fee
+		t.Fee == topicMessage.Fee &&
+		t.GasPriceWei == topicMessage.GasPriceWei
+		// TODO: Add this line when functionality supports it -> erc20Address == topicMessage.Erc20Address
```
Daniel-K-Ivanov, line 104: I think that we could add it now. The topic message contains the raw data (`ethaddress, amount, fee, gasprice, erc20address`) and the `signature`.
We could add the `erc20Address` right now and submit it as part of the topic. its just that we are not going to sign it as part of the authorisation signature yet

app/services/messages/service.go
```diff
@@ -391,13 +387,14 @@ func (ss *Service) VerifyEthereumTxAuthenticity(tm encoding.TopicMessage) (bool,
 
 	if dbTx.Amount != amount ||
```
Daniel-K-Ivanov, line 388: Add `TODO` here that we should validate the `erc20Address` as-well, once the contracts are updated

app/services/transfers/service.go
```diff
@@ -126,106 +128,132 @@ func (bs *Service) SaveRecoveredTxn(txId, amount string, asset string, m memo.Me
...
 			return
 		}
 	}
 	return onSuccess, onRevert
 }
+
+func (ts *Service) SignAuthorizationMessage(txId, ethAddress, erc20Address, amount, fee, gasPriceWei string) (string, error) {
+	authMsgHash, err := auth_message.EncodeBytesFrom(txId, ethAddress, erc20Address, amount, fee, gasPriceWei)
+	if err != nil {
+		return "", errors.New(fmt.Sprintf("Failed to encode the authorisation signature for TX ID [%s]. Error: %s", txId, err))
```
Daniel-K-Ivanov, line 250: This should be a log message imo. Errors are normally short, concise and as it turns out all lower case in Go.
I would suggest to log here the error and return `err`. The place you are calling this should only state that this exact operation failed without logging the error. That is how we are doing it in the other `services` functions.

There are 2 places from which  `SignAuthorizationMessage` is called (this service and the recovery process). Once you reformat the `ProcessTranfer` function and move the logic to the `Handler`, this function will be called only from the 2 processes -> `handler` and `recovery`

app/services/transfers/service.go
```diff
@@ -126,106 +128,132 @@ func (bs *Service) SaveRecoveredTxn(txId, amount string, asset string, m memo.Me
...
+	}
+	signatureBytes, err := ts.ethSigner.Sign(authMsgHash)
+	if err != nil {
+		return "", errors.New(fmt.Sprintf("Failed to sign the authorisation signature for TX ID [%s]. Error: %s", txId, err))
```
Daniel-K-Ivanov, line 254: This should be a log message imo

app/process/recovery/recovery.go
```diff
@@ -235,61 +234,61 @@ func (r Recovery) recoverEthereumTXMessage(tm encoding.TopicMessage) error {
 	return nil
 }
 
+type TransferMessageKey struct {
```
Daniel-K-Ivanov, line 237: I think that this could be defined as private to the package struct

app/services/transfers/service.go
```diff
@@ -126,106 +128,132 @@ func (bs *Service) SaveRecoveredTxn(txId, amount string, asset string, m memo.Me
...
 	return nil
 }
 
 // ProcessTransfer processes the transfer message by signing the required
 // authorisation signature submitting it into the required HCS Topic
-func (bs *Service) ProcessTransfer(tm encoding.TransferMessage) error {
-	authMsgHash, err := auth_message.EncodeBytesFrom(tm.TransactionId, tm.EthAddress, tm.Erc20Address, tm.Amount, tm.Fee)
+func (ts *Service) ProcessTransfer(tm encoding.TransferMessage) error {
+	gasPriceGWeiBn, err := helper.ToBigInt(tm.GasPriceGwei)
 	if err != nil {
```
Daniel-K-Ivanov, line 160: This error should be logged. If `helper.ToBigInt` returns error (as it is now), the returned error will go back to:
```Go
err = th.transfersService.ProcessTransfer(*transferMsg)
	if err != nil {
		th.logger.Errorf("Processing of TX [%s] failed", transferMsg.TransactionId)
		return
	}
```
And we won't log the exact error. 
With the current design of the services. The error is logged inside it, and the caller is not logging the exact error (since it will be logged twice).

app/services/transfers/service.go
Daniel-K-Ivanov, line 167: We do not need to log the error. It is already logged inside the `SigAuthorizationMessage`. Here we only need to state that the `SigAuthorizationMessage` operation failed for TX ID

app/services/transfers/service.go
```diff
@@ -126,106 +128,132 @@ func (bs *Service) SaveRecoveredTxn(txId, amount string, asset string, m memo.Me
...
 		tm.EthAddress,
 		tm.Amount,
 		tm.Fee,
+		gasPriceWei,
 		signature)
 
-	sigMsgBytes, err := signatureMessage.ToBytes()
+	err = ts.PrepareAndSubmitToTopic(signatureMessage)
+	if err != nil {
+		ts.logger.Errorf("Failed to Prepare and Submit TransactionID [%s] to the HCS Topic. Error [%s].", tm.TransactionId, err)
```
Daniel-K-Ivanov, line 181: I don't think that you should log the error. All of the errors are logged inside the `PrepareAndSubmitToTopic`. Here you should only state that this operation failed.

app/process/recovery/recovery.go
```diff
@@ -235,61 +234,61 @@ func (r Recovery) recoverEthereumTXMessage(tm encoding.TopicMessage) error {
...
+			EthAddress:    txnMessage.EthAddress,
+			Amount:        txnMessage.Amount,
+			Fee:           txnMessage.Fee,
+			GasPriceWei:   txnMessage.GasPriceWei,
+			Erc20Address:  txnMessage.ERC20ContractAddress,
+		}
+		signatureMessagesMap[key] = append(signatureMessagesMap[key], txnMessage.Signature)
+	}
+
+	for txn, txnSignatures := range signatureMessagesMap {
```
Daniel-K-Ivanov, line 267: nit: I would say that a better name for the variable is `transfer` or `tr` and not `txn

app/process/recovery/recovery.go
```diff
@@ -235,61 +234,61 @@ func (r Recovery) recoverEthereumTXMessage(tm encoding.TopicMessage) error {
...
-	//for _, s := range signatures {
-	//	if signature == s {
-	//		return true, nil
-	//	}
-	//}
-	return false, nil
+func (r Recovery) hasSubmittedSignature(data *TransferMessageKey, signatures []string) (bool, *encoding.TopicMessage) {
+	signature, err := r.transfers.SignAuthorizationMessage(data.TransactionId, data.EthAddress, data.Erc20Address, data.Amount, data.Fee, data.GasPriceWei)
+	if err != nil {
+		r.logger.Errorf("Failed to Validate and Sign TransactionID [%s]. Error [%s].", data.TransactionId, err)
```
Daniel-K-Ivanov, line 284: We should return the error here and handle it in the iteration above. 
If the `SignAuthorizationMessage` fails, we log the error in the `range` above and state that this `TXID` will not be recovered and you should do a `continue` (skipping this tx recovery due to invalid data as we do in the other places in the recovery

app/process/recovery/recovery.go
```diff
@@ -235,61 +234,61 @@ func (r Recovery) recoverEthereumTXMessage(tm encoding.TopicMessage) error {
...
-	//	r.logger.Errorf("Failed to Validate and Sign TransactionID [%s]. Error [%s].", data.TransactionId, err)
-	//}
-	//
-	//for _, s := range signatures {
-	//	if signature == s {
-	//		return true, nil
-	//	}
-	//}
-	return false, nil
+func (r Recovery) hasSubmittedSignature(data *TransferMessageKey, signatures []string) (bool, *encoding.TopicMessage) {
```
Daniel-K-Ivanov, line 281: This function seems little off to me. It implies that it does something very trivial and returns `true/false` but it actually signs the authorization message, checks whether it is contained or not and returns the formatted signature message otherwise. 

I think that it would be better if
1. we do the signing in the context of the `processUnfinishedOperations` function in the `range` where you iterate the `signatureMessagesMap`, 
2. call this function to check whether the signature is contained in the array. The function simply returns true/false
3. If `false` do the `encoding.NewSignatureMessage` (but again, in the scope of `processUnfinishedOperations`) and not here

app/process/recovery/recovery.go
```diff
@@ -235,61 +234,61 @@ func (r Recovery) recoverEthereumTXMessage(tm encoding.TopicMessage) error {
...
+	}
+
+	for txn, txnSignatures := range signatureMessagesMap {
+		hasSubmittedSignature, topicMessage := r.hasSubmittedSignature(txn, txnSignatures)
+
+		if !hasSubmittedSignature {
+			err = r.transfers.PrepareAndSubmitToTopic(topicMessage)
+			if err != nil {
+				r.logger.Errorf("Failed to prepare and submit Transaction ID [%s] to the HCS Topic. Error [%s].", txn.TransactionId, err)
+				return err
```
Daniel-K-Ivanov, line 274: I think that we should do `continue` rather than returning the error. If we return the error, we will break the whole recovery for the other TX and validators wouldn't be able to recover anything else (if f.e a transfer is in invalid format) and we don't want that.

See `transfersRecovery` or `topicMessagesRecovery` for reference and more precisely the `continue` in the loops

Daniel-K-Ivanov (changes requested):
We can remove the files in the `process->model` folder. We are not using them anymore

app/persistence/message/message.go
```diff
@@ -79,3 +80,25 @@ func (m Repository) GetMessagesFor(txId string) ([]TransactionMessage, error) {
...
+	var messages []*TransactionMessage
+
+	err := m.dbClient.Preload("transactions").Raw("SELECT " +
+		"transaction_messages.transaction_id, " +
+		"transaction_messages.eth_address, " +
+		"transaction_messages.amount, " +
+		"transaction_messages.fee, " +
+		"transaction_messages.signature, " +
+		"transaction_messages.erc20_contract_address, " +
+		"transaction_messages.gas_price_wei " +
```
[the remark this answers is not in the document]
georgiyazovaliiski: SELECT * will get everything from both tables, therefore will retrieve unnecessary data and will (probably) cause ambiguity in column names, making the query to stop working at all.

app/persistence/transaction/transaction.go
```diff
@@ -40,6 +42,8 @@ const (
 	StatusCompleted = "COMPLETED"
 	// StatusRecovered TODO after recovery is completed
 	StatusRecovered = "RECOVERED"
+	// StatusFailed
```
Daniel-K-Ivanov, line 45: nit: add info that TX are updated to failed once Ethereum TX reverts

app/persistence/transaction/transaction.go
```diff
@@ -308,3 +294,23 @@ func isValidStatus(status string, possibleStatuses []string) bool {
 	}
 	return false
 }
+
+func (tr *Repository) GetUnprocessedTransactions() ([]Transaction, error) {
+	var transactions []Transaction
+
+	err := tr.dbClient.Raw("SELECT " +
```
Daniel-K-Ivanov, line 301: nit: use `?` syntax in the query for parameters

app/process/handler/transfer/handler_test.go
```diff
@@ -34,6 +34,7 @@ const (
 	accountID      = "0.0.99661"
 	submissionTxID = "0.0.99661--62135596800-0"
 	signature      = "f9f9c16aa2ac71b8341d9187c37c2b8dd8152c4a27fe70f8fcf60d56456166ce704c3f1df4831d66e26879a32cb764d928de346418c1f0f116cba14d78a4dfac1b"
+	gasPriceWei    = "1000000000"
```
Daniel-K-Ivanov, line 37: unused variable

app/process/model/transaction/txid.go
```diff
@@ -22,7 +22,6 @@ import (
 	"strings"
 )
 
-// TODO not necessary -> remove
 type TxId struct {
```
Daniel-K-Ivanov, line 25: This code is used only in the e2e tests. I think that we have 2 options:
- Remove the code from here and in the e2e tests add a util function for the `FromHederaTransactionID`
- Move this code to `encoding` and use it in the `e2e` tests and the `mirror-node` client (the function equivalent to this logic is `parseIntoQueryableTx`

app/process/recovery/recovery.go
```diff
@@ -235,61 +239,68 @@ func (r Recovery) recoverEthereumTXMessage(tm encoding.TopicMessage) error {
 	return nil
 }
 
+type transferMessageKey struct {
```
Daniel-K-Ivanov, line 242: unused struct

app/process/recovery/recovery.go
```diff
@@ -235,61 +239,68 @@ func (r Recovery) recoverEthereumTXMessage(tm encoding.TopicMessage) error {
...
-	//	_, err = r.transfersService.HandleTopicSubmission(ctm, signature)
-	//	if err != nil {
-	//		return errors.New(fmt.Sprintf("Could not submit Signature [%s] to Topic [%s] - Error: [%s]", signature, r.topicID, err))
-	//	}
-	//	r.logger.Infof("Successfully Validated")
-	//}
-	//}
+	unprocessedTransactions, err := r.transactions.GetUnprocessedTransactions()
+	if err != nil {
+		r.logger.Fatalf("Failed to get all unprocessedTransactions messages. Error: %s", err)
```
Daniel-K-Ivanov, line 254: Should not be `Fatal` but `error`

app/process/recovery/recovery.go
```diff
@@ -235,61 +239,68 @@ func (r Recovery) recoverEthereumTXMessage(tm encoding.TopicMessage) error {
...
+		return err
+	}
+
+	for _, transfer := range unprocessedTransactions {
+		weiBn, err := helper.ToBigInt(transfer.GasPriceWei)
+		if err != nil {
+			r.logger.Errorf("Could not parse GasPriceWei [%s] to a Big Integer for TX [%s]. Skipping further execution", transfer.GasPriceWei, transfer.TransactionId)
```
Daniel-K-Ivanov, line 261: This error log is false. We are not `skipping further execution`. 
See the other recovery for reference:
`Skipping recovery for TX [%s]. Could not parse gas price [%s] to bigint`

app/process/recovery/recovery.go
```diff
@@ -235,61 +239,68 @@ func (r Recovery) recoverEthereumTXMessage(tm encoding.TopicMessage) error {
...
+			continue
+		}
+		gwei := ethereum.WeiToGwei(weiBn)
+
+		valid, erc20Address := r.contracts.IsValidBridgeAsset(transfer.Asset)
+		if !valid {
+			r.logger.Errorf("Specified Asset [%s] for TX ID [%s] is not supported.", transfer.Asset, transfer.TransactionId)
```
Daniel-K-Ivanov, line 268: nit: add `Skipping recovery for TX [%s]. Specified..`

app/process/recovery/recovery.go
```diff
@@ -235,61 +239,68 @@ func (r Recovery) recoverEthereumTXMessage(tm encoding.TopicMessage) error {
...
+			erc20Address,
+			transfer.Amount,
+			transfer.Fee,
+			gwei.String(),
+			transfer.ExecuteEthTransaction)
+
+		if transferMsg.ExecuteEthTransaction {
+			err = r.transfers.VerifyFee(*transferMsg)
+			if err != nil {
+				r.logger.Errorf("Fee validation failed for TX [%s]. Skipping further execution", transferMsg.TransactionId)
```
Daniel-K-Ivanov, line 284: nit: add `Skipping recovery for TX [%s]. Fee validation ..`

app/process/recovery/recovery.go
```diff
@@ -235,61 +239,68 @@ func (r Recovery) recoverEthereumTXMessage(tm encoding.TopicMessage) error {
...
-	//	r.logger.Errorf("Failed to Validate and Sign TransactionID [%s]. Error [%s].", data.TransactionId, err)
-	//}
-	//
-	//for _, s := range signatures {
-	//	if signature == s {
-	//		return true, nil
-	//	}
-	//}
-	return false, nil
+func (r Recovery) hasSubmittedSignature(signature string, signatures []string) bool {
```
Daniel-K-Ivanov, line 298: not used anymore

test/mocks/service/transfer_service_mock.go
```diff
@@ -12,6 +12,30 @@ type MockTransferService struct {
...
 
+func (mts *MockTransferService) ProcessTransfer(tm encoding.TransferMessage) error {
+	args := mts.Called(tm)
+	if args.Get(0) == nil {
+		return nil
+	}
+	return args.Get(0).(error)
+}
+
+func (mts *MockTransferService) SignAuthorizationMessage(txId, receiverAddress, erc20Address, amount, fee, gasPriceWei string) (string, error) {
```
Daniel-K-Ivanov, line 23: not used anymore?

test/mocks/service/transfer_service_mock.go
```diff
@@ -12,6 +12,30 @@ type MockTransferService struct {
...
+	args := mts.Called(txId, receiverAddress, erc20Address, amount, fee, gasPriceWei)
+	if args.Get(1) == nil {
+		return args.Get(0).(string), nil
+	}
+	return args.Get(0).(string), args.Get(1).(error)
+}
+
+func (mts *MockTransferService) SubmitAuthorizationMessage(tm encoding.TopicMessage) error {
```
Daniel-K-Ivanov, line 31: not used anymore?

app/services/transfers/service.go
```diff
@@ -70,54 +71,55 @@ func NewService(
...
-
-	verified, e := proof.Verify(tx.TransactionID, stateProof)
-	if e != nil {
-		return nil, errors.New(fmt.Sprintf("State proof verification failed. Error [%s]", e))
-	}
-
-	if !verified {
-		return nil, errors.New("State proof not valid")
-	}
+	// TODO: Uncomment when State Proof Method gets updated accordingly
```
Daniel-K-Ivanov, line 80: Is there a problem with the state proof?

Daniel-K-Ivanov (changes requested):

app/services/transfers/service.go
```diff
@@ -126,106 +128,133 @@ func (bs *Service) SaveRecoveredTxn(txId, amount string, asset string, m memo.Me
...
 		return err
 	}
 	return nil
 }
 
-// ProcessTransfer processes the transfer message by signing the required
-// authorisation signature submitting it into the required HCS Topic
-func (bs *Service) ProcessTransfer(tm encoding.TransferMessage) error {
-	authMsgHash, err := auth_message.EncodeBytesFrom(tm.TransactionId, tm.EthAddress, tm.Erc20Address, tm.Amount, tm.Fee)
+func (ts *Service) submitAuthorizationMessage(tm encoding.TopicMessage) error {
```
Daniel-K-Ivanov, line 156: It seems strange that `submitAuthorizationMessage` does:
- submit
- getting tx from db
- updating tx record
- creating callbacks
- Attaching callbacks to transaction for monitoring

I don't have a strong opinion about it, but to me it seems that we might be better off if the logic in this function is placed in `ProcessTransfer`. Not to mention that the error logs are redundant since we have a 2 layer logic here (1st layer `ProcessTransfer`, 2nd layer `submitAuthorizationMessage`

app/services/transfers/service.go
georgiyazovaliiski: The structure retrieved from the GET query is changed entirely. It needs update.

e2e/e2e_test.go
```diff
@@ -290,3 +290,31 @@ func verifyEthereumTXExecution(setup *setup.Setup, ethTransactionHash string, wh
...
+		Seconds:   split[0],
+		Nanos:     split[1],
+	}
+}
+
+func (txId hederaTxId) String() string {
+	return fmt.Sprintf("%s-%s-%s", txId.AccountId, txId.Seconds, txId.Nanos)
+}
+
+func (txId hederaTxId) Timestamp() string {
```
Daniel-K-Ivanov, line 318: nit: not used

Daniel-K-Ivanov (approved):

PR #154, opened 2021-03-24
(#137) Address-0-0 Memo E2E Test

### LINKED ISSUE #137: Implement E2E tests for 0-0 memos
Implement e2e test for sending a bridge transfer with the option for user submission of ETH TX.
- Test that HBARs can be transferred
- Test that tokens can be transferred

georgiyazovaliiski:
**Detailed description**:
An E2E Test for a scenario where the memo is in a format Address-0-0. The validators should not be validating fee or sending an ETH Transaction Execution Message. I also modified the e2e configuration file and Hedera node initialization function a bit.

Fixes #137 

**Checklist**
- [ ] Documentation added
- [x] Tests updated


failfmi (dismissed):

failfmi (dismissed):

e2e/e2e_test.go
```diff
@@ -229,27 +240,21 @@ func verifyTopicMessages(setup *setup.Setup, transactionResponse hedera.Transact
 					ethTransMsgCollected++
 					fmt.Println(fmt.Sprintf("Received Ethereum Transaction Hash [%s]", msg.GetTopicEthTransactionMessage().EthTxHash))
 				}
-
-				// Check whether we collected everything
```
Daniel-K-Ivanov, line 233: I think that leaving the blocking channel is better. We will save execution time and not wait for 60 seconds timeout. In order to make it work we just have to remove the hardcoded 1 so the if becomes:
```go
if expectedValidatorsCount == ethSignaturesCollected &&
 ethTransMsgCollected == expectedEthTxMessageCount {
	c1 <- true
}
```

e2e/e2e_test.go
```diff
@@ -188,14 +199,13 @@ func sendHbarsToBridgeAccount(setup *setup.Setup, memo string) (*hedera.Transact
 	return &res, err
 }
 
-func verifyTopicMessages(setup *setup.Setup, transactionResponse hedera.TransactionResponse, t *testing.T) string {
+func verifyTopicMessages(setup *setup.Setup, transactionResponse hedera.TransactionResponse, expectedSignaturesCount int, expectedEthTxMessageCount int, t *testing.T) string {
```
Daniel-K-Ivanov, line 202: Do we need to pass `expectedSignaturesCount`?

Daniel-K-Ivanov (changes requested):
We should update the E2E docs with the network type as-well

e2e/e2e_test.go
georgiyazovaliiski: This is needed in order to generalize the verification of different scenarios. With a format such as address-0-0 we are waiting for no ethTxMessage, therefore we pass a 0 at the assertion.

failfmi (approved):

Daniel-K-Ivanov (approved):

PR #184, opened 2021-04-02
Feature: remove watcher-sdk

### LINKED ISSUE #179: Remove the Watcher SDK

### LINKED ISSUE #180: Signature message received before actual transfer
If somehow the topic signature message is received before the actual transfer, the handling of the topic signature message will result with invalid transaction id

failfmi:
**Detailed description**:
* Remove `hedera-watcher-sdk`
* Introduce Pair - each pairs consists of watcher and handler, having a go channel as a centralized point
* Bug fixes on validation

Fixes #179
Fixes #180 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


app/process/watcher/publisher/publish.go
```diff
@@ -18,19 +18,11 @@ package publisher
...
-	message, e := proto.Marshal(m)
-	if e != nil {
-		log.Fatalf("[%s] - Failed marshalling response - ID: [%s]\n", typeMessage, id)
-	}
-
-	q.Push(&types.Message{
-		Payload: message,
-		Type:    typeMessage,
+func Publish(m proto.Message, q *pair.Queue) {
+	q.Push(&pair.Message{
```
Daniel-K-Ivanov, line 25: now that the publisher is so "thin" and only 1 line of code is the actual publishing, wouldn't it be better to remove it and call directly `q.Push` from the watchers.

app/process/watcher/transfer/watcher.go
```diff
@@ -51,28 +51,33 @@ type Watcher struct {
...
+	if err != nil {
+		log.Fatalf("Could not start Crypto Transfer Watcher for account [%s] - Error: [%s]", accountID, err)
+	}
+
 	return &Watcher{
 		transfers:        transfers,
 		client:           client,
-		accountID:        accountID,
+		accountID:        id,
 		typeMessage:      process.CryptoTransferMessageType,
```
Daniel-K-Ivanov, line 70: We do not need this anymore

Daniel-K-Ivanov (changes requested):
Now that we don't use the protobufs. Wouldn't it be better to:
- remove the `transfer_message.proto` altogether
- Change the `encoding` package to `model`
- Introduce `transfer` and `message` folders. `transfer` folder has only `transfer.go` there and `message` folder has `signature.go` and `ethereum-tx.go`. Those models are to be used by the `watcher/handlers` and the business layer of the validator such as the `transfer` and `message` services.

Example:
- transfer model has:
`NewTransfer` -> called once the watcher parses an incoming transfer and passes it to the handler. The variables populated there are all of the required data such as `txId`, `receiver`, `nativeToken` etc..
The `NewTransfer` instance can be called if a signature message is retrieved before the transfer is created, again in the watcher.

The `messages` package files could have `FromBytes` -> used by the handler to parse the incoming bytes from the protobuf.
`signature.go` will have its own `NewSignature` and `ethereum-tx.go` will have its own `NewEthereumTx` function.

Once we have those structures, the logic that we have now at `auth-message.go` can actually be added to the `transfer.go`:
- `EncodeBytesFrom` added to `transfer.go` as `GetSignableBytes`



Now that we have removed the type messages, we can delete the `process/config.go` enum and use those variables in the repository since those values are used only for the `status` repository. (p.s `HCS_CRYPTO_TRANSFER` should be actually simply `TRANSFER` and `HCS_TOPIC_MSG` could be only `TOPIC_MESSAGE`)

app/services/messages/service.go
```diff
@@ -305,6 +307,22 @@ func (ss *Service) submitEthTxTopicMessage(transferID, messageHash, ethereumTxHa
 	return ss.hederaClient.SubmitTopicConsensusMessage(ss.topicID, ethTxHashBytes)
 }
 
+func (ss *Service) awaitTransfer(transferID string) (*entity.Transfer, error) {
```
Daniel-K-Ivanov, line 310: Maybe we could add inline comment on why we do this, since this is an edge case that we discovered. Its important info imo

app/encoding/transfer-message.go
```diff
@@ -51,6 +53,14 @@ func NewTransferMessageFromBytes(data []byte) (*TransferMessage, error) {
 	return &TransferMessage{transferMsg}, nil
 }
 
+func NewTransferMessageFromInterface(data interface{}) (*TransferMessage, error) {
```
Daniel-K-Ivanov, line 56: Is it necessary to have a separate function only for the `data.(*TransferMessage)` line of code and add 4-5 lines of error handling, return statement and function definition

Daniel-K-Ivanov (changes requested):

app/persistence/status/status.go
```diff
@@ -19,11 +19,16 @@ package status
...
 	"fmt"
 	"github.com/limechain/hedera-eth-bridge-validator/app/persistence/entity"
-	"github.com/limechain/hedera-eth-bridge-validator/app/process"
 	log "github.com/sirupsen/logrus"
 	"gorm.io/gorm"
 )
 
+// Hedera Watcher SDK message types
+const (
+	TransferMessageType = "TRANSFER"
```
Daniel-K-Ivanov, line 28: Shouldn't this be only `TransferType`?

Daniel-K-Ivanov (commented):
I think that `auth-message` could be removed as well. We simply need to move the `EncodeBytesFrom` function to `transfer.go`. Better name for the function might be `GetSignableBytes`

Daniel-K-Ivanov (approved):

Daniel-K-Ivanov (approved):

PR #187, opened 2021-04-05
E2E - DB Verification

### LINKED ISSUE #75: Verify DB records after E2E execution
- Implement a DB verification service inside the `e2e` folder.
- The service must be able to easily assert that a given record exists in the DB. 
- The service could be using the already implemented repositories for the validator node

We would like to ideally be able to:
- verify that a record is created by the`TransactionRepo` and it has the correct properties at the end of the execution 
- verify that there are TransactionMessage records created by the `MessageRepo` and all of them have correct properties at the end of the execution.

georgiyazovaliiski:
**Detailed description**:
I added a database service to perform a check on records, resulting from E2E scenarios.
Modified E2E configuration by adding dbs object with three configurations (alice, bob and carol)
Added a read-only user to perform the checks.

Fixes #75 

There are two fields, currently not taken into consideration in the Equals functions for both models. This is because I did not find a suitable way to retrieve the information needed.

**Checklist**
- [ ] Documentation added
- [x] Tests updated


e2e/setup/config.go
```diff
@@ -73,10 +74,13 @@ func getConfig(config *Config, path string) error {
...
-	SenderAccount hederaSDK.AccountID
-	TopicID       hederaSDK.TopicID
-	Clients       *clients
+	BridgeAccount   hederaSDK.AccountID
+	SenderAccount   hederaSDK.AccountID
+	TopicID         hederaSDK.TopicID
+	Clients         *clients
+	DBVerifierAlice *db_validation.Service
+	DBVerifierBob   *db_validation.Service
+	DBVerifierCarol *db_validation.Service
```
Daniel-K-Ivanov: I think that it would be best to have only 1 DB validation service that will handle the validation of the all provided databases -> `DBService`

e2e/service/database/service.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/limechain/hedera-eth-bridge-validator/app/persistence"
+	"github.com/limechain/hedera-eth-bridge-validator/app/persistence/entity"
+	"github.com/limechain/hedera-eth-bridge-validator/app/persistence/message"
+	"github.com/limechain/hedera-eth-bridge-validator/app/persistence/transfer"
+	"github.com/limechain/hedera-eth-bridge-validator/config"
+	"github.com/limechain/hedera-eth-bridge-validator/e2e/model"
+	log "github.com/sirupsen/logrus"
+)
+
+type Service struct {
```
Daniel-K-Ivanov, line 18: I think that the db validation service should handle checks for all databases. This would mean that he should receive a config for the array of Db's and we simply do:
`database.VerifyTransferRecord`
`database.VerifyMessageRecords`

e2e/e2e_test.go
```diff
@@ -78,10 +84,30 @@ func Test_E2E(t *testing.T) {
...
+			status:          entity_transfer.StatusCompleted,
+			statusSignature: entity_transfer.StatusSignatureMined,
+			statusEthTx:     entity_transfer.StatusEthTxMined,
+			statusEthTxMsg:  entity_transfer.StatusEthTxMsgMined,
+		},
+		ethTransactionHash,
+		true, t)
+
+	// Step 5 - Verify Message Record in the database
+	verifyMessageRecordsInDB(setupEnv.DBVerifiers, expectedTxRecord, receivedSignatures, t)
```
Daniel-K-Ivanov, line 110: I think that it would be best if we make it cleaner and just do 1 step for both validations `verifyDatabaseRecords`

e2e/model/sig-triplet.go
```diff
@@ -0,0 +1,6 @@
+package model
```
Daniel-K-Ivanov, line 1: - The name of the file suggests it is a `triplet` but the name of the struct is `duplet`. 
- I think that `duplet` is actually `pair`

Daniel-K-Ivanov (changes requested):

app/helper/ethereum/helper.go
```diff
@@ -114,6 +114,19 @@ func GetAddressBySignature(hash []byte, signature []byte) (string, error) {
 	return crypto.PubkeyToAddress(*pubKey).String(), nil
 }
 
+func RecoverSignerFromStr(signature string, authMsgBytes []byte) (string, string, error) {
```
failfmi, line 117: Can we add names to return values for clarification what is the first and second value

failfmi (approved):

e2e/e2e_test.go
```diff
@@ -121,79 +118,58 @@ func Test_E2E_Only_Address_Memo(t *testing.T) {
...
 
-type statusInfo struct {
-	status          string
-	statusSignature string
-	statusEthTx     string
-	statusEthTxMsg  string
-}
-
-func verifyTransferRecordInDB(routerContract *router.Router, DBVerifiers []*database.Service, transactionID hedera.TransactionID, nativeToken, txFee, gasPrice string, statuses statusInfo, ethTransactionHash string, shouldExecuteEthTx bool, t *testing.T) *entity.Transfer {
+func prepareExpectedTransfer(setupEnv *setup.Setup, transactionID hedera.TransactionID, nativeToken, txFee, gasPriceWei string, statuses database.ExpectedStatuses, ethTransactionHash string, shouldExecuteEthTx bool, t *testing.T) *entity.Transfer {
```
Daniel-K-Ivanov, line 148: nit: maybe we can pass only the `setupEnv.Clients.RouterContract` instead of the whole `setup.Setup` prop

e2e/e2e_test.go
```diff
@@ -121,79 +118,58 @@ func Test_E2E_Only_Address_Memo(t *testing.T) {
...
-func verifyMessageRecordsInDB(DBVerifiers []*database.Service, record *entity.Transfer, signatures []model.SigDuplet, t *testing.T) {
-	for _, dbVerifier := range DBVerifiers {
-		exist, expectedMessageRecords, err := dbVerifier.SignatureMessagesExist(record, signatures)
-		if err != nil {
-			t.Fatalf("Could not figure out if messages for [%s] are the expected messages [%v] - Error: [%s].", record.TransactionID, expectedMessageRecords, err)
-		}
-		if !exist {
-			t.Fatalf("Messages for [%s] are not [%v].", record.TransactionID, expectedMessageRecords)
-		}
+func verifyDatabaseRecords(setupEnv *setup.Setup, expectedRecord *entity.Transfer, signatures []string, t *testing.T) {
```
Daniel-K-Ivanov, line 138: nit: maybe we could pass only the `setupEnv.DbValidation` entity instead of the whole `setupEnv`

Daniel-K-Ivanov (approved):

failfmi (approved):

Daniel-K-Ivanov (approved):

PR #193, opened 2021-04-07
Waiting for N confirmations

### LINKED ISSUE #144: Wait for N block confirmations
The current validator does not wait for `n` number of blocks before processing an Ethereum TX. In order to mitigate the risk of uncle blocks, we should have the option to specify in the environmental variables the number of blocks that the `Ethereum Watcher` should wait for before processing an TX.

georgiyazovaliiski:
**Detailed description**:
I added a WaitBlocks function in the ethereum client.

Fixes #144 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


docs/configuration.md
```diff
@@ -26,6 +26,7 @@ Name                                                                | Default
 `hedera.validator.port`                                             | 5200                                                | The port on which the application runs.
 `hedera.eth.node_url`                                               | ""                                                  | The endpoint of the Ethereum node.
 `hedera.eth.router_contract_address`                                | ""                                                  | The address of the Router contract.
+`hedera.eth.bridge_waiting_blocks`                                  | ""                                                  | The number of blocks to wait before processing an ethereum event
```
failfmi, line 29: defaulv value should be 5

failfmi: actual yaml value is hedera.eth.waiting_blocks

config/application.yml
```diff
@@ -13,6 +13,7 @@ hedera:
   eth:
     node_url:
     router_contract_address:
+    waiting_blocks: 5
```
Daniel-K-Ivanov, line 16: nit: I think that a better name is `block_confirmations`

docs/configuration.md
```diff
@@ -26,6 +26,7 @@ Name                                                                | Default
 `hedera.validator.port`                                             | 5200                                                | The port on which the application runs.
 `hedera.eth.node_url`                                               | ""                                                  | The endpoint of the Ethereum node.
 `hedera.eth.router_contract_address`                                | ""                                                  | The address of the Router contract.
+`hedera.eth.waiting_blocks`                                         | 5                                                   | The number of blocks to wait before processing an ethereum event
```
Daniel-K-Ivanov, line 29:
```suggestion
`hedera.eth.waiting_blocks`                                         | 5                                                   | The number of block confirmations to wait for before processing an ethereum event
```

app/clients/ethereum/ethereum.go
```diff
@@ -131,3 +131,35 @@ func (ec *Client) waitForTransactionReceipt(hash common.Hash) (txReceipt *types.
 
 	return ec.Client.TransactionReceipt(context.Background(), hash)
 }
+
+func (ec *Client) WaitBlocks(transactionHash string, blockNumber uint64) error {
```
Daniel-K-Ivanov, line 135: I think that it would be better if we return a channel that resolves once the transaction has more than `n` block confirmations. That way the "client" of this function will have to option to either to do blocking wait for this or continue with its logic.

app/clients/ethereum/ethereum.go
```diff
@@ -131,3 +131,35 @@ func (ec *Client) waitForTransactionReceipt(hash common.Hash) (txReceipt *types.
...
+		}
+		currentBlockNumberBn := new(big.Int).SetUint64(currentBlockNumber)
+
+		if target.Cmp(currentBlockNumberBn) <= 0 {
+			_, isPending, err := ec.Client.TransactionByHash(context.Background(), common.HexToHash(transactionHash))
+			if err != nil {
+				return err
+			}
+			if isPending {
+				err := ec.WaitBlocks(transactionHash, target.Uint64())
```
Daniel-K-Ivanov, line 155: I am not entirely sure about this one. If its pending, we must `waitForTheTransactionReceipt` -> meaning that the transaction is mined and then call `WaitBlocks` with a new `target` which is the `new transaction receipt` block

Daniel-K-Ivanov (changes requested):

app/clients/ethereum/ethereum.go
```diff
@@ -131,3 +135,31 @@ func (ec *Client) waitForTransactionReceipt(hash common.Hash) (txReceipt *types.
...
+	for {
+		currentBlockNumber, err := ec.BlockNumber(context.Background())
+		if err != nil {
+			ec.logger.Errorf("[%s] Failed retrieving block number.", raw.TxHash.String())
+			return err
+		}
+
+		if target <= currentBlockNumber {
+			receipt, err := ec.TransactionReceipt(context.Background(), raw.TxHash)
+			if err != nil {
```
Daniel-K-Ivanov, line 150: Lets parse the 2 types of errors -> if not found or any other.
If not found -> `[%s] Ethereum TX went into an uncle block` with `debug` level of the log
if any other -> `[%s] Failed to get Transaction receipt` and return `err`

app/clients/ethereum/ethereum.go
```diff
@@ -131,3 +135,31 @@ func (ec *Client) waitForTransactionReceipt(hash common.Hash) (txReceipt *types.
...
+				ec.logger.Infof("[%s] Ethereum TX went into an uncle block.", raw.TxHash.String())
+				return err
+			}
+
+			if receipt.Status == 1 {
+				ec.logger.Debugf("[%s] Transaction was successfully mined", raw.TxHash.String())
```
Daniel-K-Ivanov, line 156: nit: the transaction technically was already mined the moment we receive the event. Lets change the log to `[%] Transaction received [ec.config.BlockConfirmations] block confirmations`

app/process/watcher/ethereum/watcher.go
```diff
@@ -28,44 +29,59 @@ import (
...
 func (ew *Watcher) handleLog(eventLog *routerContract.RouterBurn, q *pair.Queue) {
-	log.Infof("New Burn Event Log for [%s], Amount [%s], Receiver Address [%s] has been found.",
+	ew.logger.Infof("New Burn Event Log for [%s], Amount [%s], Receiver Address [%s] has been found.",
 		eventLog.Account.Hex(),
 		eventLog.Amount.String(),
 		eventLog.Receiver)
-	// TODO: push to queue with message type, corresponding to ETH Handler
+
+	if eventLog.Raw.Removed {
+		ew.logger.Infof("[%s] Uncle block transaction was removed.", eventLog.Raw.TxHash)
```
Daniel-K-Ivanov, line 75: I think that it should be `debug` level

app/process/watcher/ethereum/watcher.go
```diff
@@ -28,44 +29,59 @@ import (
...
 		case eventLog := <-events:
-			ew.handleLog(eventLog, q)
+			go ew.handleLog(eventLog, q)
 		}
 	}
 }
 
 func (ew *Watcher) handleLog(eventLog *routerContract.RouterBurn, q *pair.Queue) {
-	log.Infof("New Burn Event Log for [%s], Amount [%s], Receiver Address [%s] has been found.",
+	ew.logger.Infof("New Burn Event Log for [%s], Amount [%s], Receiver Address [%s] has been found.",
```
Daniel-K-Ivanov, line 69: Maybe it would be best if we use this log after we have the block confirmations. Instead of the line 83 log. 
Here we can add log that is `debug` level just so that we know that event was received, but after the block confirmations are done we should have the normal `info` log.

Proposal for this log:
`ew.logger.Debugf("New Burn Event Log received. Waiting block confirmations")` <- we could include TX has maybe as-well

Daniel-K-Ivanov (changes requested):

Daniel-K-Ivanov (approved):

Daniel-K-Ivanov:
Lets resolve the e2e test before merging

PR #197, opened 2021-04-09
Correct Choice of Timestamp on Restart

### LINKED ISSUE #190: Watcher restarting would begin from initial timestamp
If the watcher restarts, querying transfers/messages would again begin from the initial timestamp

georgiyazovaliiski:
**Detailed description**:
I changed 2 lines of code in order to skip timestamp picking and directly continue with the last timestamp in the database on watcher restart.

Closes #190 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/app/domain/client/mirror-node.go
+++ b/app/domain/client/mirror-node.go
@@ -29,9 +29,14 @@ type MirrorNode interface {
 	GetMessagesAfterTimestamp(topicId hedera.TopicID, from int64) ([]mirror_node.Message, error)
 	// GetMessagesForTopicBetween returns all topic messages for a given topic between timestamp `from` included and `to` excluded
 	GetMessagesForTopicBetween(topicId hedera.TopicID, from, to int64) ([]mirror_node.Message, error)
-	GetAccountTransaction(transactionID string) (*mirror_node.Response, error)
+	// GetTransaction gets all data related to a specific transaction id or returns an error
+	GetTransaction(transactionID string) (*mirror_node.Response, error)
+	// GetStateProof sends a query to get the state proof. If the query is successful, the function returns the state.
+	// If the query returns a status != 200, the function returns an error.
 	GetStateProof(transactionID string) ([]byte, error)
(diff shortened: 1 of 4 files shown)


app/domain/client/mirror-node.go
```diff
@@ -32,6 +32,7 @@ type MirrorNode interface {
 	GetAccountTransaction(transactionID string) (*mirror_node.Response, error)
 	GetStateProof(transactionID string) ([]byte, error)
 	AccountExists(accountID hedera.AccountID) bool
+	TopicExists(topicID hedera.TopicID) bool
```
Daniel-K-Ivanov, line 35: nit: inline docs

app/process/watcher/message/watcher.go
```diff
@@ -60,22 +60,29 @@ func NewWatcher(client client.MirrorNode, topicID string, repository repository.
 }
 
 func (cmw Watcher) Watch(q *pair.Queue) {
+	if !cmw.client.TopicExists(cmw.topicID) {
+		cmw.logger.Errorf("Error incoming: Could not start monitoring topic [%s] - Topic not found.", cmw.topicID.String())
```
Daniel-K-Ivanov, line 64: In the error logs if we are using `Errorf` it is indicated that it is error. I think that there is no need to further emphasise the error by having "Error incoming"

app/process/watcher/transfer/watcher.go
```diff
@@ -74,27 +74,27 @@ func NewWatcher(
 }
 
 func (ctw Watcher) Watch(q *pair.Queue) {
-	accountAddress := ctw.accountID.String()
-	_, err := ctw.statusRepository.GetLastFetchedTimestamp(accountAddress)
+	if !ctw.client.AccountExists(ctw.accountID) {
+		ctw.logger.Errorf("Error incoming: Could not start monitoring account [%s] - Account not found.", ctw.accountID.String())
```
Daniel-K-Ivanov, line 78: same comment for "Error incoming"

Daniel-K-Ivanov (approved):

PR #224, opened 2021-04-20
Feature: Native fees

### LINKED ISSUE #200: Paying out fees in native asset
The contracts and validator node should be refactored such that fees should be paid out in the native currency that is being bridged. For more information refer to: 
https://github.com/LimeChain/hedera-improvement-proposal/blob/master/HIP/hip-0000-bridge.md

### LINKED ISSUE #219: E2E: Test_HBAR to submit mint transaction
Update `Test_HBAR` to:
* Submit mint transaction
* Wait for transaction
* Validate balance of `receiver`
* Reduce send amount to less (e.g. 10 hbars) - so that the config account does not go bankrupt

failfmi:
**Detailed description**:
* Distribute native fees on transfers to bridge account
* Distribute native fees on burn events of wrapped assets
* Rename `NativeToken, WrappedToken` -> `NativeAsset, WrappedAsset`

Fixes #219 
Fixes #200

**Checklist**
- [x] Documentation added
- [x] Tests updated


app/services/scheduled/service.go
```diff
@@ -0,0 +1,133 @@
...
+		return
+	}
+
+	switch txReceipt.Status {
+	case hedera.StatusIdenticalScheduleAlreadyCreated:
+		s.handleScheduleSign(id, *txReceipt.ScheduleID)
+	case hedera.StatusSuccess:
+		s.logger.Infof("[%s] - Updating db status to Submitted with TransactionID [%s].",
+			id,
+			hederahelper.ToMirrorNodeTransactionID(txReceipt.ScheduledTransactionID.String()))
```
Daniel-K-Ivanov, line 67: Do we need to log this here. Now that we have `onSuccess` and `onFail` it does not seem right for this function to know what we are going to do on success. It might be better to log it as `debug` in the `repository` `UpdateStatusSubmitted` as-well-as in the inner `updateStatus` function. Another way to do it is to log it in the definition of the callbacks in the `burn-event` service f.e

app/services/transfers/service.go
```diff
@@ -52,12 +60,21 @@ func NewService(
...
+	bridgeAccount string,
+	scheduledService service.Scheduled,
 ) *Service {
 	tID, e := hedera.TopicIDFromString(topicID)
 	if e != nil {
 		panic(fmt.Sprintf("Invalid monitoring Topic ID [%s] - Error: [%s]", topicID, e))
 	}
+	bridgeAccountID, e := hedera.AccountIDFromString(bridgeAccount)
+	if e != nil {
+		panic(fmt.Sprintf("Invalid BridgeAccountID [%s] - Error: [%s]", bridgeAccount, e))
```
Daniel-K-Ivanov, line 76: nit: Lets use `log.fatalf` for consistency

app/services/transfers/service.go
```diff
@@ -204,15 +242,106 @@ func (ts *Service) ProcessTransfer(tm model.Transfer) error {
 	return nil
 }
 
+func (ts *Service) processFeeTransfer(transferID string, feeAmount int64, nativeAsset string) {
+	transfers, err := ts.distributor.DistributeToMembers(feeAmount)
+	if err != nil {
+		ts.logger.Errorf("[%s] Fee - Failed to Distribute to Members. Error: [%s].", transferID, err)
```
Daniel-K-Ivanov, line 248: missing `return` here

config/config.go
```diff
@@ -102,6 +99,8 @@ type Hedera struct {
 	BridgeAccount string   `yaml:"bridge_account" env:"VALIDATOR_CLIENTS_HEDERA_BRIDGE_ACCOUNT"`
 	PayerAccount  string   `yaml:"payer_account" env:"VALIDATOR_CLIENTS_HEDERA_PAYER_ACCOUNT"`
 	TopicId       string   `yaml:"topic_id" env:"VALIDATOR_CLIENTS_HEDERA_TOPIC_ID"`
+	FeePercentage int64    `yaml:"fee_percentage" env:"VALIDATOR_CLIENTS_HEDERA_FEE_PERCENTAGE"`
```
Daniel-K-Ivanov, line 102: I think that if its `int64` we should be using the same concept for precision as in the contracts. If we are not going to use that concept, we need to be using float most probably in order to represent f.e fee of `1.9%`

Daniel-K-Ivanov (changes requested):

Daniel-K-Ivanov (approved):
LGTM!

PR #232, opened 2021-04-28
Feature: Sign router address

failfmi:
**Detailed description**:
* Sign router address as part of message
* Update contract wrappers to latest
* Store router address in Transfer table

Fixes #none

* Updated e2e config to be on Mumbai testnet

**Checklist**
- [ ] Documentation added
- [x] Tests updated


app/services/transfers/service.go
```diff
@@ -329,6 +331,9 @@ func (ts *Service) TransferData(txId string) (service.TransferData, error) {
 		ts.logger.Errorf("[%s] - Failed to query Transfer with messages. Error: [%s].", txId, err)
 		return service.TransferData{}, err
 	}
+	if t == nil || t.Fee.Amount == "" {
+		return service.TransferData{}, nil
+	}
```
Daniel-K-Ivanov, line 336: I think that it would be better if we return `404` in the `transfer` router if the transfer is not found

Daniel-K-Ivanov (approved):

PR #233, opened 2021-04-28
Feature(api): GET Tx ID by Event ID

failfmi:
**Detailed description**:
* Expose `/api/v1/events/:id/tx` - retrieves transaction id by event id
* E2E: validate api returns correct transaction id

Fixes #none

* After #232 

**Checklist**
- [ ] Documentation added
- [x] Tests updated


app/domain/service/burn-event.go
```diff
@@ -23,4 +23,6 @@ type BurnEvent interface {
 	// ProcessEvent processes the burn event by submitting the appropriate
 	// scheduled transaction, leaving the synchronization of the actual transfer on HCS
 	ProcessEvent(event burn_event.BurnEvent)
+	// TransactionID returns the corresponding transaction id
```
Daniel-K-Ivanov, line 26: nit: This inline doc seems confusing. Maybe this would be better:
`TransactionID returns the corresponding Scheduled Transaction paying out the fees to validators and the amount being bridged to the receiver address`

app/router/burn-event/burn-event.go
```diff
@@ -0,0 +1,40 @@
...
+		txID, err := burnService.TransactionID(eventID)
+		if err != nil {
+			render.Status(r, http.StatusInternalServerError)
+			render.JSON(w, r, response.ErrorResponse(response.ErrorInternalServerError))
+
+			logger.Errorf("Router resolved with an error. Error [%s].", err)
+			return
+		}
+
+		render.JSON(w, r, txID)
```
Daniel-K-Ivanov, line 32: Wouldn't it be better if we return 404 on not found?

Daniel-K-Ivanov (approved):

PR #240, opened 2021-05-17
Docs/integration

radtonev:
**Detailed description**:
Added documentation on how to integrate with the bridge

**Checklist**
- [X] Documentation added
- [ ] Tests updated


docs/integration.md
```diff
@@ -0,0 +1,129 @@
+﻿# Integration with Hedera <-> EVM-chain bridge
+The Bridge provides users with functionality to transfer HBAR or HTS tokens from Hedera to EVM-based chain or Wrapped HBAR and Wrapped Tokens from EVM-based chain to Hedera. The Bridge is operated by registered validators that provide signatures for every requested transfer. The transfer is processed when the majority of validators verify the transfer (supermajority).
+
+## 1. Hedera -> EVM-chain
```
Daniel-K-Ivanov, line 4: I think that this document should be focused on what exactly should be done by the integrator step-by-step.
So it will be better if we do not have 1.1 How it works, but rather ->
- Step 1. Initiate transfer of HBAR/HTS
- Step 2. Wait for confirmation
- Step 3. Claim Asset on EVM chain

The goal would be for someone to read the document and know what he has to do in order to integrate

docs/integration.md
```diff
@@ -0,0 +1,129 @@
...
+### 1.4 Service fee distribution
+
+The main incentive for the Validators is the `service fee` paid by users. The fee is a percentage of the transferred amount, paid on the native asset. The Service fee is configurable property and determined by the validators.
+
+Fees are payed out from the Bridge account. Each Validator creates a Scheduled transaction and transfers the `service fee` amount from the Bridge account to the list of validators equally. Due to the nature of Scheduled Transactions, only one will be successfully executed, creating a scheduled Entity and all others will fail with `IDENTICAL_SCHEDULE_ALREADY_CREATED` error, and the transaction receipt will include the `ScheduleID` of the first submitted transaction. All validators, except the one that successfully created the Transaction execute `ScheduleSign` and once `n out of m` validators execute the Sign operation, the transfer of the fees will be executed.
+
+## 2. EVM-chain -> Hedera
+This functionality allows the user to transfer Wrapped HBAR or any supported by the bridge Wrapped Tokens from EVM-based chain to Hedera. 
+
+### 2.1 Supported wrapped tokens
```
Daniel-K-Ivanov, line 65: IMO it would be better to rename the section to:
`2. 1. Query supported tokens`

docs/integration.md
```diff
@@ -0,0 +1,129 @@
...
+
+In order to get all the supported wrapped tokens by the bridge the user must do two things:
+
+ 1. Get the wrapped tokens count from the Router contract by calling the function `wrappedAssetsCount()`.
+ 2. Call the Router contract function `wrappedAssetAt(uint256  index)` for every value between 0 ... wrappedAssetsCount-1. Each time the function will return the address of the ERC20 wrapped token contract.
+
+If the user needs the name of the wrapped token he can retrieve it from public mapping in the Router contract by the ERC20 address:
```
Daniel-K-Ivanov, line 72: We should mention that `wrappedAssetAt` returns the `address` of the wrapped token. If the users want to see to which Hedera asset the ERC20 maps, he can do that by querying the `wrappedToNative` map. In that sense its not correct to say "if the user needs the name of the wrapped token". It should be: "In order to get the corresponding Hedera native asset, one can query the following mapping:

`mapping(address => bytes) public wrappedToNative;` 

where `address` represents the ERC20 address of the wrapped asset and `bytes` represent the HTS Entity ID or simply `HBAR` (in the case for hbars`)

It would be good to show example -> how does `bytes` look like for `HBAR`. It is a constant so we should be able to say: `0x2131221` corresponds to HBAR. We should give example for `HTS` token ID as-well

docs/integration.md
```diff
@@ -0,0 +1,129 @@
...
+
+	mapping(address => bytes) public wrappedToNative;
+
+### 2.1 How it works
```
Daniel-K-Ivanov, line 76: I think that it would be more clear for the users if its `Step 1.  Burn the Wrapped Asset`

docs/integration.md
```diff
@@ -0,0 +1,129 @@
...
+Transfer from EVM-chain to Hedera is achieved by submitting a **burn operation** to the Router Contract. 
+There are two supported contract functions by which this can be done:
+
+ - `burn(amount: uint256, receiver: bytes, wrappedAsset: address)`
+
+Argument | Description
+---------- | ----------
+**amount** | The amount of wrapped tokens to be bridged
+**receiver** | The Hedera account to receive the wrapped tokens
```
Daniel-K-Ivanov, line 85: Maybe we can show an example of how we can encode the bytes for transferring to account `0.0.1000`

docs/integration.md
```diff
@@ -0,0 +1,129 @@
...
+**wrappedAsset** | The corresponding wrapped asset contract address
+
+After the burn operation is completed a _burn_ event is fired which is captured by the validators. The event contains information about the burned amount and the receiver. After the validators capture the event they distribute the service fee and schedule a transaction to transfer the remaining amount to the receiving Hedera account.
+
+
+ - `burnWithPermit(wrappedAsset: address, receiver: bytes, amount: uint256, deadline: uint256, v, r ,s)`
```
Daniel-K-Ivanov, line 91: Maybe it would be better if we add information about the permit logic. Show a JS snippet on how to sign v,r,s and specify a `deadline`

docs/integration.md
```diff
@@ -0,0 +1,129 @@
...
+
+Argument | Description
+---------- | ----------
+**wrappedAsset** | The corresponding wrapped asset contract address
+**receiver** | The Hedera account to receive the wrapped tokens
+**amount** | The amount of wrapped tokens to be bridged
+**deadline**: | Timestamp of the deadline
+**v, r, s** | Information about the signature
+
+### 2.2 Getting information about the scheduled transaction
```
Daniel-K-Ivanov, line 101: IMO it would be better if we name it `Step 2. Monitor the Transfer` and add information on the REST API call that users can execute in order to query the scheduled transaction from the Validator Node.

Daniel-K-Ivanov (changes requested):

Daniel-K-Ivanov (approved):

PR #297, opened 2021-09-10
Feature: Recovery

### LINKED ISSUE #143: Recovery (Ethereum -> Hedera)
A recovery mechanism for Ethereum burn transactions should be implemented similar to the current Hedera -> Ethereum operations.

On validator startup, we should check whether there are any missed ethereum events for `burn`. 
If there are -> we should check whether the operation was already fulfilled by the operators (up to discussion on how exactly we will do that)

### LINKED ISSUE #285: Implement Recovery Mechanism

failfmi:
**Detailed description**:
Upon validator execution, get all fees/schedules with status Submitted and check whether they have executed successfully.
For validators depending on the provided parameter for recovery, either process past events in read-only mechanism, or directly execute them.

Fixes #285 
Fixes #143 

After #295 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


app/process/watcher/evm/watcher.go
```diff
@@ -17,47 +17,146 @@
...
+	mappings c.Assets,
+	startBlock int64,
+	validator bool) *Watcher {
+	targetBlock, err := evmClient.GetClient().BlockNumber(context.Background())
+	if err != nil {
+		log.Fatalf("Could not retrieve latest block. Error: [%s].", err)
+	}
+
+	if startBlock == 0 {
+		_, err := repository.GetLastFetchedTimestamp(contracts.Address().String())
```
Daniel-K-Ivanov, line 65: I think that it is misleading to store the blocknumber at the column for timestamp in the db. Wouldn't it be better to introduce another column for the block number?

app/process/watcher/evm/watcher.go
```diff
@@ -17,47 +17,146 @@
...
+		mappings:    mappings,
+		targetBlock: targetBlock,
+		validator:   validator,
 	}
 }
 
 func (ew *Watcher) Watch(queue qi.Queue) {
 	go ew.listenForEvents(queue)
+
+	ew.processPastLogs(queue)
```
Daniel-K-Ivanov, line 99: I think that we should first process past logs and then run the watcher. It think that if we let it the way it is, we might get into situation in which `listenForEvents` picks up an event and reacts to it and at the same time the `processPastLogs` queries for events after block X.
This way a TX may end up being processed twice

app/process/watcher/message/watcher.go
```diff
@@ -39,21 +39,48 @@ type Watcher struct {
...
+	topicID string,
+	repository repository.Status,
+	pollingInterval time.Duration,
+	startTimestamp int64) *Watcher {
 	id, err := hedera.TopicIDFromString(topicID)
 	if err != nil {
 		log.Fatalf("Could not start Consensus Topic Watcher for topic [%s] - Error: [%s]", topicID, err)
 	}
 
+	targetTimestamp := time.Now().UnixNano()
```
Daniel-K-Ivanov, line 56: I think that this property is not used anywhere

Daniel-K-Ivanov (changes requested):

app/process/watcher/evm/watcher.go
failfmi: The problem with the watch subscriptions is that they always start with the latest block. So, if you first process the past events and then start the watch subscriptions, you might end up with skipped events. Better process them twice than never. Also, most of the logic regarding that will revert in an error later on, which is not a real error - it will only be a duplication.

app/process/watcher/message/watcher.go
failfmi: `Create`

Daniel-K-Ivanov (approved):

PR #322, opened 2021-11-02
Fix: Filter EVM logs instead of subscriptions

### LINKED ISSUE #318: Validator Nodes Fatally Crash on RPC Timeout
ABOUT:
Validator nodes will fatally crash on receiving HTTP timeouts from RPC calls to the Node Urls. This issue happens both when the bridge is being used and also when no transactions are going through bridge. If there is a bad response from Infura or other RPC providers, it will cause the node to crash and requires a manual restart.

LOGS:
`ERRO[2021-10-21T03:50:43.2332205Z] Failed to subscribe for Burn Event Logs for contract address [OMITTED]. Error [websocket: bad handshake (HTTP status 502 Bad Gateway)].  context="EVM Router Watcher [OMITTED]"
FATA[2021-10-21T03:50:43.448793149Z] Failed to subscribe for WatchMemberUpdated Event Logs for contract. Error websocket: bad handshake (HTTP status 502 Bad Gateway)  context="Contract Service"
exit status 1`

Expected Behaviour:
If node receives a HTTP timeout or 502 from RPC it should not fatally exit, instead it should re-poll the RPC at a later time and remain running.

failfmi:
**Detailed description**:
* Filter EVM logs instead of subscriptions

Fixes #318 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


app/process/watcher/evm/watcher.go
```diff
@@ -85,58 +91,95 @@ func NewWatcher(
...
 
 	topics := [][]common.Hash{
 		{
 			burnHash,
 			lockHash,
+			memberUpdatedHash,
 		},
 	}
 	addresses := []common.Address{
 		ew.contracts.Address(),
```
Daniel-K-Ivanov, line 169: Are we 100% sure that events with the same signature, emitted in a `router` TX will be ignored

app/process/watcher/evm/watcher.go
```diff
@@ -85,58 +91,95 @@ func NewWatcher(
...
 	}
+
 	query := &ethereum.FilterQuery{
 		FromBlock: new(big.Int).SetInt64(fromBlock),
+		ToBlock:   new(big.Int).SetInt64(endBlock),
 		Addresses: addresses,
 		Topics:    topics,
 	}
+
 	logs, err := ew.evmClient.FilterLogs(context.Background(), *query)
```
Daniel-K-Ivanov, line 179: FYI: If there are more than 1000 events matching this `FilterLog` the node will reject the query and will throw `>400` status

app/process/watcher/evm/watcher.go
```diff
@@ -85,58 +91,95 @@ func NewWatcher(
...
+		if fromBlock > toBlock {
+			time.Sleep(ew.sleepDuration)
+			continue
+		}
+
+		err = ew.processLogs(fromBlock, toBlock, queue)
+		if err != nil {
+			ew.logger.Errorf("Failed to process logs. Error: [%s].", err)
+			time.Sleep(ew.sleepDuration)
+			continue
```
Daniel-K-Ivanov, line 143: nit: we do not need those 2 lines of code

Daniel-K-Ivanov (changes requested):
- In evm/watcher.go -> line 317 you need to update the `Debugf` not to state that we are waiting for block confirmations

app/clients/evm/client.go
```diff
@@ -84,14 +84,16 @@ func (ec *Client) ValidateContractDeployedAt(contractAddress string) (*common.Ad
 }
 
 // GetBlockTimestamp retrieves the timestamp of the given block
-func (ec *Client) GetBlockTimestamp(blockNumber *big.Int) (uint64, error) {
+func (ec *Client) GetBlockTimestamp(blockNumber *big.Int) uint64 {
```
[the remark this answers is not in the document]
failfmi: I think it would better to hang and retry than to end the processing of a given transfer, just because something has timed out

app/process/watcher/evm/watcher.go
```diff
@@ -85,58 +91,95 @@ func NewWatcher(
...
 
-func (ew Watcher) processPastLogs(queue qi.Queue) {
+func (ew Watcher) beginWatching(queue qi.Queue) {
 	fromBlock, err := ew.repository.Get(ew.contracts.Address().String())
 	if err != nil {
 		ew.logger.Errorf("Failed to retrieve EVM Watcher Status fromBlock. Error: [%s]", err)
-		return
 	}
 
 	ew.logger.Infof("Processing events from [%d]", fromBlock)
```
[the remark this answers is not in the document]
failfmi: 1. yes
2. yes,  it used to be fatalf

app/process/watcher/evm/watcher.go
failfmi: That's why you are filtering by `addresses`. They are used to restrict matches to events created by those specific addresses.

app/process/watcher/evm/watcher.go
failfmi: I've left them on purpose, so that if the error is timeout related, it will pause for a bit, rather than trying again immediately

app/process/watcher/evm/watcher.go
failfmi: The limitation should based in the querying node. I guess we will adapt depending on the node config.

app/process/watcher/evm/watcher.go
Daniel-K-Ivanov: Aren't we going to pause for a bit even if we remove those 2 lines because of line 197?

Daniel-K-Ivanov (dismissed):

Daniel-K-Ivanov (approved):

PR #328, opened 2021-11-09
Fix(evm): Split logs filtering into chunks

failfmi:
**Detailed description**:
* Most EVM nodes have a limit on how many blocks can be queried via `eth_getLogs`. It would be best to have a max logs blocks, so that in case of a big query, erroneous behaviour will not occur and the query will be split into chunks of N queries.
* For example, a max logs blocks of `500`. Suppose that we have to process a range of `4200` blocks. This will be split into 9 chunks, 8 of which will be `500` and one `200`.

Fixes #none

**Checklist**
- [x] Documentation added
- [ ] Tests updated


app/process/watcher/evm/watcher.go
```diff
@@ -55,7 +55,15 @@ type Watcher struct {
 	filterConfig  FilterConfig
 }
 
-var defaultSleepDuration = 15 * time.Second
+// Certain node providers (Alchemy, Infura) have a limitation on how many blocks
+// eth_getLogs can process at once. For this to be mitigated, a maximum amount of blocks
+// is introduces, splitting the request into chunks with a range of N.
```
Daniel-K-Ivanov, line 60:
```suggestion
// is introduced, splitting the request into chunks with a range of N.
```

Daniel-K-Ivanov (dismissed):

Daniel-K-Ivanov (approved):

PR #348, opened 2021-12-20
feat(metrics): validators' participation rate

svetlio8:
**Detailed description**:
Dashboard: add Prometheus and Grafana to the initial setup. Implement participation rate as a metric in the dashboard.

**Checklist**
- [x] Documentation added
- [x] Tests updated


app/process/handler/message/handler.go
```diff
@@ -47,12 +53,16 @@ func NewHandler(
 		log.Fatalf("Invalid topic id: [%v]", topicId)
 	}
 
+	participationRate := prometheusServices.NewGaugeMetric("validators_participation_rate", "Participation rate: Track validators' activity in %.")
```
failfmi, line 56: Can we extract this metric information either in a config file or in a constants file that will store all future metrics

svetlio8: done with constants

failfmi (approved):

Issue #360, opened 2022-01-10
Operator account balance metric


svetlio8:
To add also the Operator account balance as a metric

svetlio8:
Done in https://github.com/LimeChain/hashport-validator/pull/368

Issue #354, opened 2022-01-06
Minted tokens


svetlio8:
The number of minted wrapped tokens on the network.

svetlio8:
Done in https://github.com/LimeChain/hashport-validator/pull/369

Issue #353, opened 2022-01-06
Locked Tokens


svetlio8:
The number of locked native tokens in the account.

svetlio8:
Done in https://github.com/LimeChain/hashport-validator/pull/369

PR #369, opened 2022-01-17
Add operator account balance as a metric. Assets balance, total suppl…

svetlio8:
Add operator account balance as a metric. Assets balance, total supply, count, and the Bridge Account balance as metrics.

Signed-off-by: svetlio <svetlin.madzharov@gmail.com>

Issues: #353 #354 #360 

**Checklist**
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/app/services/prometheus/service_test.go
+++ b/app/services/prometheus/service_test.go
@@ -6,25 +6,19 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/stretchr/testify/assert"
 	"testing"
-	"time"
 )
 
 var (
-	gauge           prometheus.Gauge
-	service         *Service
-	gaugeName       = "GaugeName"
-	gaugeHelp       = "GaugeHelp"
-	pollingInterval = 5 * time.Second
+	gauge     prometheus.Gauge
+	service   *Service
+	gaugeName = "GaugeName"
+	gaugeHelp = "GaugeHelp"
 )
 
 func Test_New(t *testing.T) {
 	setup()
 
-	actualService := NewService(
-		mocks.MHederaMirrorClient,
-		mocks.MTransferRepository,
-		pollingInterval,
-	)
+	actualService := NewService()
 
 	assert.Equal(t, service, actualService)
 }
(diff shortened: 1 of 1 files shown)


nnedkov-limechain (approved):

app/process/watcher/prometheus/watcher.go
```diff
@@ -73,37 +103,256 @@ func (pw Watcher) Watch(q qi.Queue) {
...
+		constants.AssetMetricHelpSuffix,
+		address)
+	pw.initAndRegAssetMetric(asset, pw.countAssetsMetrics, metricName, help)
+}
+
+func (pw Watcher) tokenIDtoMetricName(id string) string {
+	replace := strings.Replace(id, constants.DotSymbol, constants.ReplaceDotSymbol, constants.DotSymbolRep)
+	result := fmt.Sprintf("%s%s", constants.AssetMetricsNamePrefix, replace)
+	return result
+}
```
failfmi, line 211: This does not require to be associated to the struct

failfmi (changes requested):
* Monitoring config is not parsed successfully.
* Crashes with nil pointer
```
prometheus.Watcher.setBridgeAccAssetsMetrics(
```

app/process/watcher/prometheus/watcher.go
svetlio8: Updated

app/process/watcher/prometheus/watcher.go
```diff
@@ -73,37 +104,259 @@ func (pw Watcher) Watch(q qi.Queue) {
...
+
+func (pw Watcher) setEvmCountMetric(routerInstance *router.Router, asset string) {
+	countMetric := pw.prometheusService.GetGauge(pw.countAssetsMetrics[asset])
+	count, e := routerInstance.NativeTokensCount(&bind.CallOpts{})
+	if e != nil {
+		panic(e)
+	}
+	pw.setEvmAssetMetric(countMetric, asset, "Count of", count)
+}
```
failfmi, line 356: I think it would be better to clarify what we are counting -> native tokens, in the logs and in the metrics

failfmi: In the docs as well

failfmi (approved):

app/process/watcher/prometheus/watcher.go
svetlio8: fixed in https://github.com/LimeChain/hashport-validator/pull/381

PR #399, opened 2022-02-14
feat: NFT support

failfmi:
**Detailed description**:
* Update abis with latest contracts
* Config: split tokens into fungible & nft
* Config: introduce fees for nft tokens
* Support NFT messages for EVM minting
* Listen for BurnERC721 event and execute a scheduled nft transaction to receiver
* Update Topic messages format
* E2E test for Native NFT Hedera -> Wrapped NFT EVM
* E2E test for Wrapped EVM NFT -> Native Hedera

Fixes #<issue number>

Tests will fail until the configuration is updated

**Checklist**
- [x] Documentation added
- [ ] Tests updated


app/clients/hedera/mirror-node/model/transaction.go
```diff
@@ -65,44 +72,104 @@ type (
...
 }
 
-// GetIncomingTransfer returns the token amount OR the hbar amount that is credited to the specified
-// account for the given transaction. It depends on getIncomingAmountFor() and getIncomingTokenAmountFor()
-func (t Transaction) GetIncomingTransfer(account string) (string, string, error) {
+// GetIncomingTransfer returns the transfer to an account in the following order:
+// 1. Checks if there is an NFT transfer
+// 2. Checks if there is a Fungible Token transfer
+// 3. Checks if there is an HBAR transfer
+func (t Transaction) GetIncomingTransfer(account string) (isNft bool, amountOrSerialNum int64, asset string, err error) {
```
nnedkov-limechain, line 156: The return result could be packed as struct type (in ex: TrasnferResponse), as otherwise, it is too lengthy for the return values.

app/model/message/message.go
```diff
@@ -74,10 +93,26 @@ func NewSignature(sourceChainId, targetChainId uint64, transferID, asset, receiv
 		Amount:        amount,
 		Signature:     signature,
 	}
-	return &Message{topicMsg}
+	return &Message{TopicMessage: &model.TopicMessage{Message: &model.TopicMessage_FungibleSignatureMessage{FungibleSignatureMessage: topicMsg}}}
+}
+
+// NewNftSignature instantiates Signature Message struct ready for submission to the Bridge Topic
+func NewNftSignature(sourceChainId, targetChainId uint64, transferID, asset string, tokenId uint64, metadata, recipient, signature string) *Message {
```
nnedkov-limechain, line 100: You can just accept `model.TopicEthNftSignatureMessage` here as this way the signature will expect just 1 object instead of each field separate and if something changes in the struct's fields it won't need to update the parameters.

app/process/handler/message-submission/handler_test.go
```diff
@@ -68,21 +68,25 @@ var (
...
+		TargetChainId: uint64(transferRecord.TargetChainID),
+		TransferID:    transferRecord.TransactionID,
+		Asset:         transferRecord.NativeAsset,
+		Recipient:     transferRecord.Receiver,
+		Amount:        transferRecord.Amount,
+		Signature:     "signature",
 	}
-	date = time.Date(2001, time.June, 1, 1, 1, 1, 1, time.UTC)
-	txId = &hedera.TransactionID{
+	//m = &message.Message{
```
nnedkov-limechain, line 80: Do we need this to be committed?

app/services/read-only/service.go
```diff
@@ -117,6 +119,63 @@ func (s Service) FindAssetTransfer(
...
+				}
+				for _, tx := range scheduledTx.Transactions {
+					if tx.Result == hedera.StatusSuccess.String() {
+						scheduleID, err := s.mirrorNode.GetSchedule(tx.EntityId)
+						if err != nil {
+							s.logger.Errorf("[%s] - Failed to get scheduled entity [%s]. Error: [%s]", transferID, scheduleID, err)
+							break
+						}
+						if scheduleID.Memo == transferID {
+							isFound = true
```
nnedkov-limechain, line 152: You can move the logic if it is found here to skip 1 check. But also it can get too indented so not sure if it is better in terms of readability.

nnedkov-limechain (approved):

svetlio8 (approved):

nnedkov-limechain (approved):

PR #411, opened 2022-02-25
Cache chainID - set it on start

### LINKED ISSUE #405: Cache chainId instead of fetching it every time
Currently,
```go
var chain *big.Int
chain, e := ew.evmClient.ChainID(context.Background())
if e != nil {
  ew.logger.Errorf("[%s] - Failed to retrieve chain ID.", eventLog.Raw.TxHash)
  return
}
```
is called on every EVM related transfer, when this is data that can be cached upon validator start. It is not expected to change. Suggestion would be to cache it in`evmClient`.

svetlio8:
Signed-off-by: svetlio <svetlin.madzharov@gmail.com>

**Detailed description**:
Cache chainID - set it on start

Fixes #405

Not fully finished, there is still an issue with the tests.

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/app/process/watcher/evm/watcher.go
+++ b/app/process/watcher/evm/watcher.go
@@ -17,7 +17,6 @@
 package evm
 
 import (
-	"context"
 	"errors"
 	"fmt"
 	"github.com/ethereum/go-ethereum"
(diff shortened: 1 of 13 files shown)


app/process/watcher/evm/watcher.go
```diff
@@ -346,20 +340,15 @@ func (ew *Watcher) handleBurnLog(eventLog *router.RouterBurn, q qi.Queue) {
 		return
 	}
 
-	var chain *big.Int
-	chain, e := ew.evmClient.ChainID(context.Background())
-	if e != nil {
-		ew.logger.Errorf("[%s] - Failed to retrieve chain ID.", eventLog.Raw.TxHash)
-		return
-	}
+	chain := ew.evmClient.ChainID()
```
failfmi, line 343: can we rename this to `nativeChainId`

app/process/watcher/evm/watcher.go
```diff
@@ -565,13 +548,8 @@ func (ew *Watcher) handleBurnERC721(eventLog *router.RouterBurnERC721, q qi.Queu
...
 	}
 
-	var chain *big.Int
-	chain, e := ew.evmClient.ChainID(context.Background())
-	if e != nil {
-		ew.logger.Errorf("[%s] - Failed to retrieve chain ID.", eventLog.Raw.TxHash)
-		return
-	}
-	nativeAsset := ew.mappings.WrappedToNative(eventLog.WrappedToken.String(), chain.Uint64())
+	chain := ew.evmClient.ChainID()
```
failfmi, line 551: can we rename this to `sourceChainId` as well

app/process/watcher/evm/watcher.go
svetlio8: will be fixed in a new PR https://github.com/LimeChain/hashport-validator/pull/412

PR #624, opened 2022-05-20
Feature/scripts for updating bridge config

### LINKED ISSUE #601: Update the bridge config from HCS topics messages
Research how to update the config on message based on #593
- Get the config from the topic. Because of the topic message submission max size = 1024 bytes [link](https://docs.hedera.com/guides/docs/hedera-api/consensus-service/consensussubmitmessage#consensusmessagechunkinfo) - we will need to split the config into chunks. Example the `bridge.yml` config on [testnet](https://testnet.mirrornode.hedera.com/api/v1/topics/0.0.34726830/messages) is split into 5 chunks. We will need to assemble the 5 chunks and parse the config.
- Update once a new config is sent. 
- Implement the logic

Psykepro:
**Detailed description**:

Fixes #601

**Checklist**
- [x] Documentation added
- [x] Tests updated


bridge.yml
```diff
@@ -1,22 +1,24 @@
...
+      tokens:
+        fungible:
+          "HBAR":
+            fee_percentage: 10000 # 10.000%
+            coin_gecko_id: "hedera-hashgraph"
+            coin_market_cap_id: "4642"
+            min_fee_amount_in_usd: 0.001 # USD
+            networks:
+              3: "0xb083879B1e10C8476802016CB12cd2F22a896571"
+              80001: "0xb083879B1e10C14761010161B12cd2F25a896691"
```
failfmi, line 24: Can we not commit this?

Psykepro: ops yep :D  it was for test data

app/process/handler/read-only/nft/fee/handler.go
```diff
@@ -74,9 +77,15 @@ func NewHandler(
 		readOnlyService:    readOnlyService,
 		hederaNftFees:      hederaNftFees,
 	}
+
+	event.On(constants.EventBridgeConfigUpdate, event.ListenerFunc(func(e event.Event) error {
+		return bridgeCfgEventHandler(e, instance)
+	}), constants.HandlerEventPriority)
+
+	return instance
```
failfmi, line 85: This handler should not use `hederaNftFees` at all.
Line `validFee := fmh.distributor.ValidAmount(fmh.hederaNftFees[transferMsg.SourceAsset])`
Should be resolved to `validFee := fmh.distributor.ValidAmount(transferMsg.Fee)`

failfmi: This way you will not need to have an event listener.

config/parser/node.go
```diff
@@ -22,12 +22,13 @@ import "time"
...
-	Port       string     `yaml:"port"`
-	Validator  bool       `yaml:"validator"`
-	Monitoring Monitoring `yaml:"monitoring"`
+	Database            Database   `yaml:"database"`
+	Clients             Clients    `yaml:"clients"`
+	LogLevel            string     `yaml:"log_level"`
+	Port                string     `yaml:"port"`
+	Validator           bool       `yaml:"validator"`
+	Monitoring          Monitoring `yaml:"monitoring"`
+	BridgeConfigTopicId Monitoring `yaml:"bridge_config_topic_id"`
```
failfmi, line 31: This is not needed

docs/configuration.md
```diff
@@ -16,41 +16,46 @@ Unless you need to set a non-default value, it is recommended to only populate o
...
 
 Configuration for `config/bridge.yml`:
 
 | Name                                                          | Default | Description                                                                                                                                                                                                                                                          |
 |---------------------------------------------------------------|---------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `bridge.use_local_config`                                     | false   | If `use_local_config` is `false` it keep using the local config, if it is `true` it will fetch the bridge config from the topics and will start watcher to keep it updated.                                                                                          |
```
failfmi, line 56: If `use_local_config` is `true` it keeps using the local config, if it is `false` it will fetch the bridge config from the topics and will start watcher to keep it updated.

bootstrap/server.go
```diff
@@ -73,25 +74,38 @@ func InitializeServerPairs(server *server.Server, services *Services, repositori
...
 }
 
-func registerTransferWatcher(server *server.Server, services *Services, repositories *Repositories, clients *Clients, configuration config.Config) {
+func registerBridgeConfigWatcher(s *server.Server, services *Services, configuration *config.Config, useLocalConfig bool, bridgeCfgTopicId hedera.TopicID) {
+	if !useLocalConfig {
+		dashboardPolling := configuration.Node.Monitoring.DashboardPolling * time.Minute
+		log.Infoln("Dashboard Polling interval: ", dashboardPolling)
+		s.AddWatcher(bridge_config.NewWatcher(services.BridgeConfig, bridgeCfgTopicId))
+	} else {
+		log.Infoln("Monitoring is disabled. No metrics will be added.")
```
failfmi, line 88: nit: log text

app/process/watcher/bridge-config/watcher.go
```diff
@@ -0,0 +1,65 @@
...
+	parsedBridge, err := w.svc.ProcessLatestConfig(w.topicID)
+	if parsedBridge != nil {
+		if parsedBridge.PollingInterval != w.pollingInterval {
+			w.pollingInterval = parsedBridge.PollingInterval
+		}
+	}
+
+	if err != nil {
+		w.logger.Errorf(err.Error())
+	}
```
failfmi, line 64: nit: check first for the error

failfmi (changes requested):

vectorman1 (dismissed):

app/router/config-bridge/config-bridge.go
```diff
@@ -24,19 +24,21 @@ import (
 )
 
 var (
-	Route = "/config/bridge"
+	Route        = "/config/bridge"
+	BridgeConfig *parser.Bridge
```
failfmi, line 28: This always returns the exact same from the initial load from the local configuration

failfmi: Actually what's happening is it fetches the latest config from the topic, saves it, then you register the router and it overrides it.

app/router/assets/assets.go
```diff
@@ -27,7 +27,8 @@ import (
 )
 
 var (
-	Route = "/assets"
+	Route        = "/assets"
+	BridgeConfig *parser.Bridge
```
failfmi, line 31: This is no longer needed, you can just use the passed argument in `New...`

app/router/config-bridge/config-bridge.go
failfmi, line 28: This is no longer needed, you can just use the passed argument in `New...`

failfmi (changes requested):

constants/events.go
```diff
@@ -0,0 +1,29 @@
...
+import "github.com/gookit/event"
+
+const (
+	EventBridgeConfigUpdate          = "config.bridge.update"
+	BridgeConfigUpdateEventParamsKey = "params"
+	ClientsEventPriority             = event.Max
+	AssetServicePriority             = event.High
+	ServiceEventPriority             = event.AboveNormal
+	WatcherEventPriority             = event.Normal
+	HandlerEventPriority             = event.BelowNormal
```
svetlio8, line 28: Are we using the `HandlerEventPriority` const somewhere?

Psykepro: No, but I left it there to be easy for the person who will need to implement an event to know which priority it needs to be.

failfmi (approved):

PR #689, opened 2022-08-09
Added a rule for a health check alert (#630)

### LINKED ISSUE #630: Add notification based on health-check metric
Since we already have Alerts configuration for Participation rate #472, it will be good to have such notification based on the health-check metric. We can have different indications based on the number of inactive validators.

Notes: Keep in mind the current configuration of the rules.yml and be careful with the single/double quotes issue.

rokn:
Signed-off-by: Antonio Mindov <antonio.mindov@limechain.tech>

**Detailed description**:
Adding a rule to Prometheus which will send alerts if the healthy validators fall under a given threshold

Also fixed problem with escaping double quotes in the Prometheus `rules.yml` because of `$value`. This was caused by the wrong pair of quotes being used in the GitHub workflow.

Fixes #630 

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/examples/three-validators/monitoring/prometheus/rules.yml
+++ b/examples/three-validators/monitoring/prometheus/rules.yml
@@ -7,8 +7,8 @@
 #        for: 1m
 #        # Labels - additional labels to be attached to the alert
 #        labels:
-#          severity: 'minor'
-#          group: 'account_amounts'
+#          severity: "minor"
+#          group: "account_amounts"
 #        annotations:
 #          description: "Fee account amount: {{ $value }} HBAR"
 #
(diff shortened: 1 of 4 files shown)


examples/three-validators/monitoring/prometheus/rules.yml
```diff
@@ -35,4 +35,16 @@
 #          group: 'validators'
 #          repeat_interval: 'long'
 #        annotations:
-#          description: "Participation Rate: {{ $value }} %"
\ No newline at end of file
+#          description: 'Participation Rate: {{ $value }} %'
+#
+#      - alert: LowHealthyValidators
+#        # Condition for alerting
+#        expr: count_validators_alive < 6
```
svetlio8, line 42: I think the example is with 3 validators and maybe 1 read-only. Also can you please check for different types of alerts? I guess 3 levels of alerting will be fine (one per validator).

rokn: Should be okay now

svetlio8 (dismissed):

Psykepro (dismissed):

vectorman1 (dismissed):

svetlio8 (approved):

PR #732, opened 2022-09-08
Adding logs for Node account IDs for which a scheduled transaction is…

### LINKED ISSUE #724: Add logs and further research the INVALID_NODE_ACCOUNT error
We should log which node is wrong so we can pinpoint if a single node is causing problems.

Psykepro:
… being signed. And on which node was executed when there is an error.

Signed-off-by: Nikolay Nedkov <nikolay.nedkov@limechain.tech>

**Detailed description**:

Fixes #724

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/app/clients/hedera/client.go
+++ b/app/clients/hedera/client.go
@@ -29,12 +29,13 @@ import (
 type Node struct {
 	client   *hedera.Client
 	maxRetry int
+	logger   *log.Entry
 }
 
 // NewNodeClient creates new instance of hedera.Client based on the provided client configuration
-func NewNodeClient(config config.Hedera) *Node {
+func NewNodeClient(cfg config.Hedera) *Node {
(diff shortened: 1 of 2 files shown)


rokn:
Can you add the log to `ExecuteScheduledMintTransaction` and `ExecuteScheduledBurnTransaction` as the error was happening there as well

Psykepro:
> Can you add the log to `ExecuteScheduledMintTransaction` and `ExecuteScheduledBurnTransaction` as the error was happening there as well

The log which is in Hedera Node Client will work for all transactions. The scheduled service is using it in the end to send the sign.

svetlio8 (approved):

trayanm (approved):

rokn (approved):

PR #766, opened 2022-11-11
Add more Logs

### LINKED ISSUE #760: Add more logs for transaction at Node Account in the tx services
We need to cover the https://github.com/LimeChain/hashport-validator/blob/main/app/services/scheduled/service.go with more logs related to the printing of Hedera nodes that are hit, not only on the failer. For example https://github.com/LimeChain/hashport-validator/blob/main/app/services/scheduled/service.go#L219
Please check also the app/clients/hedera/client.go

Coiling-Dragon:
Signed-off-by: Stanislav Viyachev <stanislav.viyachev@limechain.tech>

**Detailed description**:

Fixes #760

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/app/clients/hedera/client.go
+++ b/app/clients/hedera/client.go
@@ -135,7 +135,7 @@ func (hc Node) SubmitTopicConsensusMessage(topicId hedera.TopicID, message []byt
 		return nil, err
 	}
 
-	hc.logger.Debugf("Submit Schedule Sign, with transaction ID [%s] and Node Account IDs: %v",
+	hc.logger.Debugf("Submit Topic Consensus Message, with transaction ID [%s] and Node Account IDs: %v",
 		tx.GetTransactionID(),
 		tx.GetNodeAccountIDs(),
 	)
@@ -145,7 +145,7 @@ func (hc Node) SubmitTopicConsensusMessage(topicId hedera.TopicID, message []byt
 	if err != nil {
 		return nil, err
 	}
-	
+
 	_, err = hc.checkTransactionReceipt(response)
 
 	return &response.TransactionID, err
@@ -249,7 +249,7 @@ func (hc Node) submitScheduledTransaction(signedTransaction hedera.ITransaction,
 	scheduledTx = scheduledTx.
 		SetPayerAccountID(payerAccountID).
 		SetScheduleMemo(memo)
-	
+
 	response, err := scheduledTx.Execute(hc.GetClient())
 	return &response, err
 }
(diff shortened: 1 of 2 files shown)


rokn (commented):
I don't think we should overburden the schedule service to log info about the transactions. IMO this should be done just in the Hedera client.
Last time we missed those two from the logs, which I thought was the reason for this task:
`func (hc Node) SubmitTopicConsensusMessage(topicId hedera.TopicID, message []byte) (*hedera.TransactionID, error)`
`func (hc Node) SubmitScheduleSign(scheduleID hedera.ScheduleID) (*hedera.TransactionResponse, error)`

I checked them and currently it would require a small logic change to be able to log the node account before executing the transaction, as it's currently directly executing them without freezing. I'm not sure about changing the logic just to be able to log those things. @svetlio8 @trayanm WDYT?

svetlio8 (approved):

app/clients/hedera/client.go
```diff
@@ -125,28 +125,49 @@ func (hc Node) SubmitScheduledTokenBurnTransaction(tokenID hedera.TokenID, amoun
...
 		SetMaxRetry(hc.maxRetry).
-		Execute(hc.client)
+		FreezeWith(hc.client)
 
 	if err != nil {
 		return nil, err
 	}
 
-	_, err = hc.checkTransactionReceipt(txResponse)
+	hc.logger.Debugf("Submit Schedule Sign, with transaction ID [%s] and Node Account IDs: %v",
```
rokn, line 138:
```suggestion
	hc.logger.Debugf("Submit Topic Consensus Message, with transaction ID [%s] and Node Account IDs: %v",
```

app/clients/hedera/client.go
```diff
@@ -228,7 +249,7 @@ func (hc Node) submitScheduledTransaction(signedTransaction hedera.ITransaction,
 	scheduledTx = scheduledTx.
 		SetPayerAccountID(payerAccountID).
 		SetScheduleMemo(memo)
-
+
```
rokn, line 252: nit: whitespace

app/services/scheduled/service.go
```diff
@@ -39,9 +39,14 @@ func New(
...
-	transfers []transfer.Hedera,
-	onExecutionSuccess func(transactionID, scheduleID string), onExecutionFail, onSuccess, onFail func(transactionID string)) {
+		id,
+		nativeAsset string,
+		transfers []transfer.Hedera,
+		onExecutionSuccess func(transactionID, scheduleID string),
+		onExecutionFail,
+		onSuccess, 
+		onFail func(transactionID string),
+	) {
```
rokn, line 49: nit: This is a formatting only can you revert it?

trayanm (dismissed):

app/clients/hedera/client.go
Coiling-Dragon: done

rokn (approved):

PR #780, opened 2022-11-28
Added fix for INVALID_NODE_ACCOUNT

rokn:
**Detailed description**:
Add a temporary fix for INVALID_NODE_ACCOUNT while the SDK gets fixed.

**Which issue(s) this PR fixes**:
Fixes #779 

**Special notes for your reviewer**:

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

PR #783, opened 2022-11-30
Added fix for INVALID_NODE_ACCOUNT (#780)

rokn:
Signed-off-by: Antonio Mindov <antonio.mindov@limechain.tech>

**Detailed description**:

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/app/clients/hedera/client.go
+++ b/app/clients/hedera/client.go
@@ -18,7 +18,6 @@ package hedera
 
 import (
 	"fmt"
-
 	"github.com/hashgraph/hedera-sdk-go/v2"
 	"github.com/limechain/hedera-eth-bridge-validator/app/model/transfer"
 	"github.com/limechain/hedera-eth-bridge-validator/config"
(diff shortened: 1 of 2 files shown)


app/clients/hedera/client.go
```diff
@@ -125,50 +124,69 @@ func (hc Node) SubmitScheduledTokenBurnTransaction(tokenID hedera.TokenID, amoun
...
+		}
 
-	hc.logger.Debugf("Submit Topic Consensus Message, with transaction ID [%s] and Node Account IDs: %v",
-		tx.GetTransactionID(),
-		tx.GetNodeAccountIDs(),
-	)
+		hc.logger.Debugf("Submit Topic Consensus Message, with transaction ID [%s] and Node Account IDs: %v",
+			tx.GetTransactionID(),
+			tx.GetNodeAccountIDs(),
+		)
```
trayanm, line 143: nit: I think we need to log the number of retry, otherwise we may have the same logged line, which could be confusing

app/clients/hedera/client.go
```diff
@@ -125,50 +124,69 @@ func (hc Node) SubmitScheduledTokenBurnTransaction(tokenID hedera.TokenID, amoun
...
+		}
 
-	_, err = hc.checkTransactionReceipt(response)
+		_, err = hc.checkTransactionReceipt(response)
 
-	return &response.TransactionID, err
+		return &response.TransactionID, err
+	}
+
+	return nil, fmt.Errorf("Too many retries to submit topic message to %s", topicId.String())
```
trayanm, line 159: I believe this should be logged with `hc.logger.Debugf(...)`

trayanm (approved):

svetlio8 (approved):
Broken build & e2e, I guess it is from the Hedera SDK update.

app/clients/hedera/client.go
rokn: I definitely think this should be an error as we have an unfinished transaction.

PR #864, opened 2023-03-07
fix minAmount not defaulting

Coiling-Dragon:
**Detailed description**:
Fixes the minAmount problem for tokens with "min_fee_amount_in_usd" were defaulting to 0 instead.

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated


svetlio8 (commented):
Do you think that we do not need the same check also in `updateHbarPrice` method?

svetlio8 (approved):

PR #884, opened 2023-04-06
fixes for treshold keys and reuse old privateKeys

Coiling-Dragon:
**Detailed description**:

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/scripts/bridge/setup/cmd/setup.go
+++ b/scripts/bridge/setup/cmd/setup.go
@@ -22,7 +22,6 @@ import (
 	"github.com/limechain/hedera-eth-bridge-validator/scripts/bridge/setup"
 )
 
-var balance = hedera.NewHbar(100)
 
 func main() {
 	privateKey := flag.String("privateKey", "0x0", "Hedera Private Key")
@@ -32,7 +31,10 @@ func main() {
 	adminKey := flag.String("adminKey", "", "The admin key")
 	topicThreshold := flag.Uint("topicThreshold", 1, "Topic member keys sign threshold")
 	flag.Parse()
-	result := setup.Deploy(privateKey, accountID, adminKey, network, members, topicThreshold)
+
+	// pass empty array of private keys since we are using this script for new accounts
+	hederaPrivateKeys := make([]hedera.PrivateKey, 0)
+	result := setup.Deploy(privateKey, accountID, adminKey, network, members, hederaPrivateKeys, topicThreshold)
 	if result.Error != nil {
 		panic(result.Error)
 	}
--- a/scripts/bridge/setup/from-config/cmd/setup-from-config.go
+++ b/scripts/bridge/setup/from-config/cmd/setup-from-config.go
@@ -59,16 +59,16 @@ func main() {
 
 	prKeysSlice := strings.Split(*memberPrivateKeys, ",")
 	var hederaPrivateKeys []hedera.PrivateKey
-	for i := 0; i < len(prKeysSlice); i++ {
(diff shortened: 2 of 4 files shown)


scripts/bridge/setup/from-config/cmd/setup-from-config.go
```diff
@@ -50,20 +50,35 @@ func main() {
...
 	configPath := flag.String("configPath", "scripts/bridge/setup/extend-config/extended-bridge.yml", "Path to the 'bridge.yaml' config file")
 	flag.Parse()
 
+	prKeysSlice := strings.Split(*memberPrivateKeys, ",")
+	var hederaPrivateKeys []hedera.PrivateKey
+	for i := 0; i < len(prKeysSlice); i++ {
+		// element [0] can be empty string if the user does not provide any private keys
+		if prKeysSlice[0] == "" {
+			break
+		}
```
rokn, line 66: I think this requires some mind gymnastics to understand, so we should either:
1. Surround the for with the if instead
2. Extract this logic in a function which returns the private keys array from the string and do:
```golang
if prKeysSlice[0] == "" {
	return nil
}
```

scripts/bridge/setup/setup.go
```diff
@@ -37,9 +38,11 @@ type DeployResult struct {
 	Error              error
 }
 
-func Deploy(privateKey *string, accountID *string, adminKey *string, network *string, members *int, topicThreshold *uint) DeployResult {
+func Deploy(privateKey *string, accountID *string, adminKey *string, network *string, members *int, cachedPrivateKeys []hedera.PrivateKey, topicThreshold *uint) DeployResult {
```
rokn, line 41: I think `oldPrivateKeys` or `previousPrivateKeys` conveys a bit more clearer what those keys are

scripts/bridge/setup/setup.go
```diff
@@ -37,9 +38,11 @@ type DeployResult struct {
...
 	result := DeployResult{}
 
+	fmt.Printf("%v", cachedPrivateKeys)
```
rokn, line 44: nit: If it's not leftover add some more info

scripts/bridge/setup/cmd/setup.go
```diff
@@ -22,7 +22,6 @@ import (
 	"github.com/limechain/hedera-eth-bridge-validator/scripts/bridge/setup"
 )
 
-var balance = hedera.NewHbar(100)
```
rokn, line 25: Wasn't this used somewhere?

Coiling-Dragon: unused variable

scripts/bridge/setup/setup.go
Coiling-Dragon: forgot to delete it

scripts/bridge/setup/from-config/cmd/setup-from-config.go
Coiling-Dragon: Noted and refactored to:
```
	// element [0] can be empty string if the user does not provide any private keys
	if prKeysSlice[0] != "" {
		for i := 0; i < len(prKeysSlice); i++ {
			privateKeyFromStr, err := hedera.PrivateKeyFromString(prKeysSlice[i])
			if err != nil {
				panic(err)
			}
			hederaPrivateKeys = append(hederaPrivateKeys, privateKeyFromStr)
		}
	}
```

rokn (approved):

PR #909, opened 2023-08-15
Fixed a memory leak in the Retry functionality for EVM clients

rokn:
**Detailed description**:
The retry loop was leaking go routine from the flow which was not finished(either the timeout or the `executionFunction` because when one finished the other couldn't send a message on it's channel.

The solution is to just use a context with a timeout and leave the handling to the user which works perfectly in our case as everywhere we are using this we are already using a context.

**Checklist**
- [x] Documentation added
- [x] Tests updated

### DIFF
--- a/app/domain/service/retry.go
+++ b/app/domain/service/retry.go
@@ -17,8 +17,8 @@
 package service
 
 import (
+	"context"
 	"errors"
-	"fmt"
 	"github.com/limechain/hedera-eth-bridge-validator/app/model/retry"
 	log "github.com/sirupsen/logrus"
 	"time"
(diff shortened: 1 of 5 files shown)


app/domain/service/retry.go
```diff
@@ -28,26 +28,7 @@ const (
...
-		time.Sleep(sleepPeriod)
-		r <- retry.Result{
-			Value: nil,
-			Error: timeoutError,
-		}
-	}()
-
-	return r
-}
+var ()
```
asenslime, line 31: delete?

app/domain/service/retry.go
```diff
@@ -57,33 +38,28 @@ func timeout() <-chan retry.Result {
...
-	retryFunction = func() (interface{}, error) {
-		var executionResult retry.Result
-		select {
-		case executionResult = <-timeout():
-		case executionResult = <-executionFunction():
-		}
+	for {
+		ctx, cancel := context.WithTimeout(context.Background(), sleepPeriod)
+		executionResult := executionFunction(ctx)
+		cancel()
```
asenslime, line 47: defer cancel()?

Later, if someone uses ctx, it will be canceled and the request could not be fired

asenslime (approved):

Coiling-Dragon (approved):

Coiling-Dragon (approved):

PR #924, opened 2025-03-28
refactor: use the built-in max/min to simplify the code

pkucode:
**Detailed description**:

Use the built-in max/min from the standard library in Go 1.21 to simplify the code.  https://pkg.go.dev/builtin@go1.21.0#max

Fixes #<issue number>

**Checklist**
- [ ] Documentation added
- [ ] Tests updated

### DIFF
--- a/app/helper/big-numbers/helper.go
+++ b/app/helper/big-numbers/helper.go
@@ -30,11 +30,3 @@ func ToBigInt(value string) (*big.Int, error) {
 
 	return amount, nil
 }
-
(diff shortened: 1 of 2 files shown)


nickmandev:
Thank you for the contribution but we decided that we wont merge it for the time being. Your change increased the Go version from 1.20 ~> 1.21 which will require a lot of changes in our infra. 

Until we are ready with the version upgrade we will keep this PR open.

PR #929, opened 2026-03-26
Implement GasPriceClient

andreygLime:
Added a GasPrice client: a client.GasPrice interface and a gas_price.Client that delegates to the existing EVM stack via SuggestGasPrice, returning the suggested price in wei. Includes unit tests and a MockGasPriceClient.

### DIFF
--- a/app/clients/evm/gas-price/client.go
+++ b/app/clients/evm/gas-price/client.go
@@ -26,16 +26,22 @@ type gasPricer interface {
 	SuggestGasPrice(ctx context.Context) (*big.Int, error)
 }
 
-type Client struct {
+type GasClienter interface {
+	GetGasPrice(ctx context.Context) (*big.Int, error)
+}
+
+var _ GasClienter = (*GasClient)(nil)
+
+type GasClient struct {
 	evmClient gasPricer
 }
 
-func NewClient(evmClient gasPricer) *Client {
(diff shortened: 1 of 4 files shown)


app/clients/evm/gas-price/client.go
```diff
@@ -0,0 +1,45 @@
...
+	"context"
+	"fmt"
+	"math/big"
+)
+
+type gasPricer interface {
+	SuggestGasPrice(ctx context.Context) (*big.Int, error)
+}
+
+type Client struct {
```
vlady-kotsev: Maybe rename to GasClient

app/clients/evm/gas-price/client.go
```diff
@@ -0,0 +1,45 @@
...
+	evmClient gasPricer
+}
+
+func NewClient(evmClient gasPricer) *Client {
```
vlady-kotsev: Can we make `func NewClient(evmClient gasPricer) GasClienter` to have compile time interface check ?

app/domain/client/gas_price.go
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package client
+
+import (
+	"context"
+	"math/big"
+)
+
+type GasPrice interface {
```
vlady-kotsev: Maybe rename to GasClienter

